Issue #458, opened 2014-10-26
Ouroboros API handles upload collisions incorrectly


Diftraku:
Due to an oversight while coding the image upload, the logic to check for the collision handling is in the code but will never be reached if the first hash check (supplied by API consumer) finds a duplicate. This was put in to minimize processing of data to some extent but instead, it simply ignores all dupes regardless of configuration option if the API consumer supplies an md5 hash along the metadata.

Quick fix is to just not send this but I'm working on a patch and having it tested on my local testbed.

Diftraku:
This took a while to get done to, at least it got tested well on my local testbed :3

Diftraku:
Fixed in PR #475 and merged to develop branch in commit 67fc50e6c3c0019122bb16db8e1c55708cddd27e

PR #473, opened 2015-01-05
Fix duplicate handling

Diftraku:
Now with 17% less fat:
* Moved validation logic into OuroborosPost for sanity
* Added sanity checks for OuroborosPost values
* Changed defaults to be more sane (mainly tags)
* OuroborosPost now accepts an MD5 hash along the metadata (like it
should!)
* Fixed dupe handling logic

See issue #458 for more details.


Diftraku:
Wrong branch in the master repo, changing and creating new ->

PR #479, opened 2015-02-02
Change generated links to be aware of HTTPS.

jgen:
This changes the make_http function to be aware of HTTPS.

Note: I have not tested this change as I currently do not have a site with HTTPS/SSL enabled.

### DIFF
--- a/core/util.inc.php
+++ b/core/util.inc.php
@@ -402,10 +402,18 @@ function modify_url($url, $changes) {
  * @return string
  */
 function make_http(/*string*/ $link) {
-	if(strpos($link, "ttp://") > 0) return $link;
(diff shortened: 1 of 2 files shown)


shish:
I don't have a https site either, but the code looks good~

I wonder what support for protocol-less URLs is like now; ie "//foo.com" rather than "http://foo.com" to imply "use whatever the current protocol is". I know that works in some contexts (eg CSS and JS links in the header), but not sure about others...

muhkuh2005:
Got an HTTPS Site and it seems to work fine so far.

Yes // is a generic syntax as far as I know.
I dont know about browsers compatibility though. But it should work on all browsers in all tags.

Issue #478, opened 2015-01-31
Bulk Uploading Tag Aliases in CSV Does Not Change Tags In Affected Posts


The-Excel:
When I upload a set of tag aliases in the bulk upload form, the aliases are added to the list, but the tags in the images themselves are not changed as they are when an alias is entered manually. This causes posts with the aliased tags to become unsearchable as searching for the tag returns the alias (with zero results) instead of the old tag, but the posts themselves still have the old tag and thus cannot be searched by the new tag.

Running commit 09d03d9036d1f9562f9f86f83114f265324e4870, most recent stable release as of January 31, 2015.

Issue #470, opened 2014-12-24
Shimmie places GET parameters to form action attribute


xzfc:
E.g., there are

```
<form action="./index.php?q=/source_history/1136" method="GET">
    <input value="View Source History" type="submit">
</form>
```

In this case, user directed to `./index.php`, not `./index.php?q=/source_history/1136`.

Correct form is

```
<form action="./index.php" method="GET">
    <input type="hidden" name="q" value="/source_history/1136">
    <input value="View Source History" type="submit">
</form>
```

shish:
This got done

PR #482, opened 2015-03-21
Fix for #478 & #470 + Various other bugfixes

DakuTree:


DakuTree:
>Some databases will consider it an error if you try to insert a long string into a short field, so I think we still need to limit to whatever the field length is

Ah, hadn't thought about that. Majority of filesystems have a limit of 255 characters anyway (including extension), with [very few](http://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits) allowing more than that. Guess it would be good to limit just in case though.

>I have a vague memory of this restriction happening for performance reasons -- even when all filenames were short, even /allowing/ up to 255 made the image table slow.

Not sure if it's the best solution, but I'm guessing we could just stick the filenames in a separate table & only grab when needed?

DakuTree:
I've went ahead and done an alternate fix over @ 9fdd781b3e4842ab1086d4a48cd1a55cc55252ee which should solve the performance issues. Basically using the above idea, filenames are stored in a separate table & only grabbed when needed (which is extremely rare anyway).

The ideal fix would be to just avoid grabbing the filename column, but that isn't really possible since we won't know exactly what columns exist and what doesn't (since extensions can add columns etc.).

shish:
I've cherry-picked a bunch of these commits individually, so this pull-request is outdated and conflicting; closing this now, feel free to open another with only the unmerged changes~

Issue #490, opened 2015-06-01
Picture Bugs


PronMaster:
These are the small bugs I found with the latest version of shimmie the fit picture to width and height is bugged also the information for the picture like the source and tags box don't load it says at set if you could look into this shimmie crew this would be great thanks.

jgen:
Thanks for opening the issue.

Do you happen to have a screenshot of how the images aren't looking right?
(Or perhaps a link? -- you can email me directly if you'd rather not post here.)

Also, when you say that the source and tag box don't load, what exactly do you mean "it says at set" ?
(Sorry, I don't quite understand.)

Thank you!

PronMaster:
I sent you a email jgen

jgen:
Thanks PronMaster.

I have created a pull request to fix the issue with the "Fit to ..." not working. (See #492 )

PR #492, opened 2015-06-02
Fix issue with "Fit to Width" and other "Fit to .." options not working.

jgen:
This should fix the issue with "Fit to Width" and other "Fit to .." options not working (issue #490).

The issue here is that the zoom function is defined on a global scope in JavaScript, so if you have any other libs that also define a zoom function, Shimmie's is overridden. (Yay, JavaScript...)


PronMaster:
jgen you guys are great this fixed my issue no need to downgrade thanks a lot.

shish:
Looks good, thank you ^_^

PR #495, opened 2015-06-10
Fix MySQL error

Xetanai:
This error is caused by a trailing comma in line 20.


jgen:
While this could be merged in safely, the recent change by @shish (see commit 10e7dd2 ) fixes it for all create_table calls. :+1:

PR #506, opened 2015-08-12
Fixes #505 & other small fixes

### LINKED ISSUE #505: No handler could be found for the page 'bulk_add'
Hi! 
I installed shimmie on my Debian home-server yesterday. 
Everything was running fine - I even tested by uploading 9k~ of images using the 'Bulk Add' extension, and it worked fine.
Today, whenever I try to use the Bulk Add just like before, I get the message

`No handler could be found for the page 'bulk_add'`

Any ideas on what could it be the problem? I'll post the configuration or anything you need, ask away.
Thanks

DakuTree:
This should fix the issue in #505.
Bulk_Add no longer gets stuck in loop when checking for subdirectory & properly notifies the user if files don't exist in directory or directory doesn't exist, rather than showing no handler exist error.

Two other minor fixes included:
* Negative aliases should properly convert each aliased tag to negative, rather than just the first tag.
* Pools index link was throwing JS errors as was also being used as ID, using specified ID fixes this.

### DIFF
--- a/core/imageboard.pack.php
+++ b/core/imageboard.pack.php
@@ -1168,6 +1168,7 @@ class Tag {
 	 */
 	public static function resolve_alias($tag) {
 		assert('is_string($tag)');
+		global $database;
 
 		$negative = false;
 		if(!empty($tag) && ($tag[0] == '-')) {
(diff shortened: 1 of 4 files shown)


shish:
All looks good, merging and thank you ^_^

PR #519, opened 2015-09-25
Adding support for apache 2.4

terokorp:
This requires mod_version installed, but many distributions ship it by default

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,7 +3,14 @@
 </IfModule>
 
 <FilesMatch "\.(sqlite|sdb|s3db|db)$">
-	Deny from all
+	# Apache 2.4
+	<IfVersion >= 2.4>
+		Require all denied
+	</IfVersion>
+	# Apache 2.2
+	<IfVersion < 2.4>
+		Deny from all
+	</IfVersion>
 </FilesMatch>
 
 <IfModule mod_rewrite.c>
--- a/.htaccess
+++ b/.htaccess
@@ -3,14 +3,12 @@
 </IfModule>
 
 <FilesMatch "\.(sqlite|sdb|s3db|db)$">
-	# Apache 2.4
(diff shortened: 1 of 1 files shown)


JarJak:
but many doesn't... apache 2.4 works fine with Deny from all

terokorp:
http://httpd.apache.org/docs/2.4/mod/mod_access_compat.html
"The directives provided by mod_access_compat have been deprecated by the new authz refactoring. Please see mod_authz_host."

HungryFeline:
No, it doesn't anymore. Apache 2.4.16: ```[...] [core:alert] [...] [...] /srv/www/htdocs/.htaccess: Invalid command 'Deny', perhaps misspelled or defined by a module not included in the server configuration``` and even if it would still work, it's officially obsolete. See https://httpd.apache.org/docs/2.4/upgrading.html#run-time

You could use IfModule. If mod_version is loaded use your code, if not, just the new directives for Apache 2.4. But that could be bad for the performance.

terokorp:
yea, I got that same `Invalid command 'Deny'` and thats why made this patch.
As I wrote to patch message, "Many distributions ship mod_version by default". At first I think to use IfModule but using IfVersion was recommended to use somewhere in apache documets or (comments in there) so I decided to choise mod_version

JarJak:
Why not use both IfModule and IfVersion?

HungryFeline:
That's what I meant with
> If mod_version is loaded use your code

terokorp:
ok, I replaced it with IfModule. using IfModule and ifVersion dosent make sense.

if mod_authz_host or mod_access_compat dosen't have loaded on apache 2.3+ server gives 500 error which is better than giving permission to download database files.
And on 2.2 or lower mod_access have to be loaded or it gives 500 error.

JarJak:
+1 but it needs to te added in readme

PR #523, opened 2015-10-08
Move tag sanitization, alias resolving & tag parsing from set_tags > TagSetEvent

DakuTree:
This should help any extension which uses onTagSet. Proper tags are now sent, and metatags are not.

Other things:
* Minor speed tweaks, which is mainly avoiding looping over possible duplicate tags & doing a simple pre-check to see if a tag could be a metatag (rather than running the tag through 10~ preg_match).
* resolve_aliases should now use Tag::explode instead of explode. This fixes a rare case where an empty tag could be passed.

### DIFF
--- a/core/imageboard.pack.php
+++ b/core/imageboard.pack.php
@@ -627,8 +627,9 @@ class Image {
 	 * Send list of metatags to be parsed.
 	 *
 	 * @param [] $metatags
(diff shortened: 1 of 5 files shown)


shish:
All looks good, sorry for not noticing this earlier >_>;;;

PR #524, opened 2015-10-14
Fix the notes extension to work properly on SQLite

DanielOaks:
By default, the notes extension doesn't work with an SQLite database. Throws an error about `LAST_INSERT_ID()` not existing when you try to add a note. This PR just makes it use the equivalent SQLite function there.

It looks like Postgres would probably need something different again there. I'm not familiar with it though, so I just left it alone.

### DIFF
--- a/ext/notes/main.php
+++ b/ext/notes/main.php
@@ -276,8 +276,7 @@ class Notes extends Extension {
 				(?, ?, ?, ?, now(), ?, ?, ?, ?, ?)",
 				array(1, $imageID, $user_id, $_SERVER['REMOTE_ADDR'], $noteX1, $noteY1, $noteHeight, $noteWidth, $noteText));
 
-		$result = $database->get_row("SELECT LAST_INSERT_ID() AS noteID", array());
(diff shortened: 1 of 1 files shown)


shish:
There should be a function in the Database class (core/database.class.php) which does this for all the databases; could you use that? (Sorry for not doing that myself, I'm on a mobile network at the moment ^^)

DanielOaks:
All good. It's late here, so I'll get to it tomorrow after work or so. Thanks for the quick response!

JarJak:
Why you won't just use http://php.net/manual/en/pdo.lastinsertid.php ?

shish:
IIRC Database does use that, but wraps it, because PHP (even though it's one function, you need to call it in different ways depending on which database you use ~_~)

DanielOaks:
Alright, this seems to work fine for me. Is this watcha meant?

I don't have a Postgres setup to test the sequence name, but it matches the `table_column_seq` format that the other calls use, and that I've read is what Postgres expects.

shish:
That looks great, merging and thank you :)

Issue #508, opened 2015-09-10
Pay for New theme?


i2plover:
Is it possible for me to pay a designer (or rather is a designer who will) make a new theme for Shimmie2 with material design? So that it's up to date, and also mobile friendly at last?

I would suggest using materializecss.com

I'm willing to throw into the pot $20

shish:
I don't know any designers myself, but I would be happy for that to happen, and can assist with the backend bits

i2plover:
I can try to do it myself if I knew how to get around the shimmie2 structure. It's a bit weird from what I saw inside the php files. 

So yes, any help to guide me in the right direction would be awesome! :+1:

i2plover:
How can I edit the navigation links/the html blocks (left, right)?
How would I move the sidebar from the left to the right? 
How would I edit the "home page" look (the one with the counter on the front) 

Really, I'm kinda asking for where would I find the place to edit to do these things. Like what Php files should I edit to accomplish these things?

shish:
The main place to start would be to copy "themes/warm" to eg "themes/material" (using "warm" as an example because it shows how to customise - if you started with default then you'd only see how to leave things as default); then in that directory

- layout.class.php contains the overall layout (header, footer, navigation, main page elements)
- themelet.class.php contains various common elements (look at `core/basethemelet` to see all the functions that you can override)
- you can customise specific extensions by creating `themes/material/extension_name.theme.php` and editing there (eg if you want to customise `ext/home/theme.php`'s HomeTheme, you would make `themes/material/home.theme.php` which contains `class CustomHomeTheme extends HomeTheme {...}`)

The general idea with shimmie is that each extension will supply a set of "blocks" (a title, a bunch of HTML to show, and hints as to where on the page it belongs) then the `layout.class.php` will take care of arranging the blocks.

Editing the navigation would be in each extension

Switching sidebar from left to right would probably be a combination of changing the HTML in `layout.class.php` and the CSS in `style.css`

The home page with the counter is the "home" extension, whose theme can be customised by creating `themes/material/home.theme.php` and putting in there a class which extends the default `ext/home/theme.php`

i2plover:
Ugh.. Gonna take some time If I have to make it myself..
I'm still willing to pay someone else to do it, and it's open if you can! :+1: Post here if interested. 
Again, read first post to learn more.

dali99:
[...]

PR #527, opened 2015-10-26
Material Design Lite Theme

CrandellWS:
This is at least a base starting point for https://github.com/shish/shimmie2/issues/508

Still could use some work


shish:
Looking good for something for people to work on - can you resubmit for merging into develop rather than master?

PR #529, opened 2015-10-26
Hotfix block id character stripping

CrandellWS:
Example problem is ext/blotter uses:

  $page->add_block(
     new Block(
     "Welcome to the Blotter Editor!", <-- SEE --> !
      $html, "main", 10));


shish:
Nice catch, thanks /o/

PR #534, opened 2015-12-04
Subquery optimization on MySQL < 5.6

DakuTree:
This speeds up bulk_add when running MySQL < 5.6 (From around 5min > 3s for a batch of 200~ images).
The tags metatag should be a bit more usable now as well (although it still has issues with queries with a lot of possible results, I.E: "tags:>1").

### DIFF
--- a/core/imageboard.pack.php
+++ b/core/imageboard.pack.php
@@ -560,9 +560,22 @@ class Image {
 	 */
 	public function delete_tags_from_image() {
 		global $database;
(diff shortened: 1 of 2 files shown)


shish:
Looks good, and I'm going to assume that since unit tests pass then it doesn't break under postgres or sqlalchemy (if it does, my fault for not writing enough tests :P ). Thank you ^_^

PR #543, opened 2016-03-21
Fixes for bulk add results

sanmadjack:
The bulk add screen was having errors resulting from the block handling of the page. This allows the results to display on my system.

### DIFF
--- a/ext/bulk_add/theme.php
+++ b/ext/bulk_add/theme.php
@@ -12,9 +12,9 @@ class BulkAddTheme extends Themelet {
 		$page->add_block(new NavBlock());
 		$html = "";
 		foreach($this->messages as $block) {
-			$html .= "<br/>" . html_escape($html);
+			$html .= "<br/>" . $block->body;
 		}
-		$page->add_block(new Block("Results", $block));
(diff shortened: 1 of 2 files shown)


ext/bulk_add/theme.php
```diff
@@ -12,9 +12,9 @@ public function display_upload_results(Page $page) {
 		$page->add_block(new NavBlock());
 		$html = "";
 		foreach($this->messages as $block) {
-			$html .= "<br/>" . html_escape($html);
+			$html .= "<br/>" . $block->body;
```
shish: I'm pretty sure this still needs to be escaped - it would be pretty tricky to exploit (you'd need to trick an admin into bulk-uploading a directory where one of the file names contained an XSS payload), but I don't think that would be impossible...

sanmadjack: Sure, I can make another pull request to add it back in.

sanmadjack: html_escapeing at this location causes the HTML being sent as part of the error messages to get messed up. Fortunately the parts that generate the error message html_escape the paths when they create the messages, so I think this is okay.

PR #546, opened 2016-05-11
Fix relationships ext checking for TRUE/FALSE instead of Y/N

DakuTree:
This was causing posts to not properly show if they had a child post or not.


shish:
I think this depends on the database used - mysql uses "Y" / "N", postgres uses true / false, etc D:

DakuTree:
Ahh, didn't even think about that. Also forgot we had bool_escape to handle this :\
Updated the PR with a fix.

shish:
Looks good, will merge when travis says ok :) (And sorry for taking so long, still catching up with everything after taking a holiday x_x)

PR #548, opened 2016-05-22
Fix wildcard tag bugs

DakuTree:
This fixes two wildcard tag bugs:
* Fixes single-tag wildcard searches returning duplicate results if the image has more than one match.
* Fixes multi-tag wildcard searches not returning the correct results if wildcard tag has more than one match on the image.


shish:
Merging because this takes things from "everything is broken" to "mysql is fixed" \o/  Then will start looking at the other DBs...

PR #550, opened 2016-06-06
Have contact_link support http: and mailto:

shish:
some people have mailto: already, and some want non-mail contact links


shish:
gah wrong root

PR #552, opened 2016-06-06
Search tests

shish:
Adding some more tests so that we can see that #548 has fixed the issues


shish:
Merging while broken because it's a legit bug to fix, and hopefully the other pull request should fix it...

PR #553, opened 2016-06-16
Set ffmpeg's "overwrite output files" switch (-y)

HungryFeline:
Regenerate thumbnail doesn't work since the output file already exists and ffmpeg expects the user to decide wether to replace the file or not. With the -y switch set, ffmpeg does so without asking.

https://ffmpeg.org/ffmpeg.html#Main-options


shish:
Nice catch, thanks :D

PR #554, opened 2016-06-17
migrate to a non-deprecated (and dependency-free) cookie library

shish:
all of shimmie's JS libraries are horrifically out of date D:


DakuTree:
Looks good. Travis error seems to be due to the search tests not working..

As for the rest of shimmie's outdated libs, I've been [poking](https://github.com/DakuTree/shimmie2/tree/composer) the idea of using composer to handle library management. This also fixes most of the outdated libs (the only exception being getID3). Seems to work [pretty well](https://travis-ci.org/DakuTree/shimmie2/jobs/131601911).
Wonder if it would be worth pulling this to the main repo?

Issue #374, opened 2014-02-11
Update Shimmie version number?


jgen:
Can we update the internal version number to say something other than "trunk"?

See: https://github.com/shish/shimmie2/blob/master/core/sys_config.inc.php#L36

This would be useful trying to determine what is causing a particular issue (ie: #373)

Just an idea. :)

PR #555, opened 2016-06-18
Composer

DakuTree:
This adds composer support to shimmie, as well as fixes a bunch of related bugs.

There is two main reasons behind adding this:
* The main one being that vendor libs are barely ever being updated. We have a few libs that don't even work anymore (reCaptcha for example). Using composer makes it a lot easier to update libs, as well as keep track of which ones we actually have.  
* Another reason, is this stops people from using the "Download ZIP" version of shimmie for production, rather than using the ones on the releases page. (This was brought up [two years ago](https://github.com/shish/shimmie2/issues/374#issuecomment-34741476), and is part of the reason we have the master/develop branches now).  

Due to the nature of the PR, this also fixes a bunch of related things which is the reason it's pretty huge. I've tried to list any notable things below.
* Libs & core css/js files now generate separate files, rather than bundling everything in one. Libs are usually a lot bigger and rarely update, while core files tend to update often. Re-DL'ing every lib for a minor core change is probably not great for bandwidth.
* No longer cache CSS/JS from disabled exts.
* Moved from jquery.autocomplete to [tag-it](http://aehlke.github.io/tag-it/). This works a lot better and is much better for mobile users.
* Autocomplete is now loaded via an extension (which means it can be disabled).
* All libs have been updated to recent versions.
* reCaptcha works again.
* Moved a bunch of ext-specific libs to their own folder (these are NOT installed with composer).
* Replaced jQuery-ui with a dropin of the only function we actually use (highlight).
* image/jpeg is longer forced for media types (audio/video etc.).
* [MediaElement.js](http://mediaelementjs.com/) is now used instead of Jaris for SWF/FLV & <video> fallback.
* <video> element is no longer bigger than the parent div (big videos used to stretch outside the visible page.
* handle_video will default to ffmpeg if it can find it in the PATH on the initial ext startup.
* Cleaned up the notes ext + fixed bugs. Sadly this still doesn't really work too well.
* Added seperate development instructions to the README (for composer install etc.).
* Cleaned up install.php. Added an error if composer folder doesn't exist (this works well from deterring people from installing shimmie from the "Download ZIP" version).
* Got rid of a bunch of old decapricated HTML elements that were still be used (font, plaintext etc.).
* Cleaned up lib.
* Cleaned up travis.yml. Checking against 7.0 instead of nightly now since 7.0 is out.
* Travis now works with composer. vendor libs are cached where possible.
* Nearly all third-party js/css/swf libs are now copied into lib/vendor by composer. This helps separate any shimmie specific libs from the third-party ones.
* There is three things which aren't installed by composer. getID3 is ext-specific but is used by multiple extensions (handle_video, handle_mp3). Modernizr is custom-generated, so we can't grab that. Everything in ext/ we shouldn't be handling libs for unless it's a CORE_EXT.

If this does get pulled, it would probably be worth doing a release as well. Our last release was over a year ago (and it's actually hidden since it was only tagged, and not marked as release..).


install.php
```diff
@@ -77,8 +77,7 @@
 				</p>
 			</div>
 		</div>
-		<div style="display: none;">
-			<PLAINTEXT>
```
shish: The main point of <plaintext> here was to stop html processing entirely, but if that's not a thing any more and `display:none` is the best we can get, <pre> seems a good a tag as any

shish:
Mostly looking awesome, and that takes a few things off my own to-do list too :D

One issue with only loading JS/CSS from enabled extensions -- if we use the latest modification timestamp as the cache unique ID, then enabling an extension from the past won't change the ID. Maybe we can have ID=max(asset modification times, data/config/extensions.conf.php modification time)?

Looks like somewhere in here, "data" type responses got broken by an extra newline appearing at the beginning... (which can break JSON responses etc, since the content-length is set to the JSON length, and thus `[1,2,3]` becomes `\n[1,2,3`). Investigating that now, and then will try letting this run on my own sites for a few days to shake out any corner-case issues that came from upgrading libraries.

ext/notes/main.php
```diff
@@ -1,3 +1,4 @@
+
```
shish: excess whitespace

shish:
Actually, looks like including the config modification time into the asset modification time was already a thing, ignore that suggestion :)

ext/autocomplete/style.css
```diff
@@ -0,0 +1,8 @@
+#Navigationleft .blockbody { overflow: visible; }
+
+.tagit { background: white !important; border: 1px solid grey !important; cursor: text; }
+.tagit-choice { cursor: initial; }
+input[name=search] ~ input[type=submit] { display: block !important; }
```
shish: Why `display:block`? `inline-block` seems nicer on the things I'm looking at (eg, on the home page, block causes the input box and the submit button to be on different lines, with the submit button floating off against one edge of the screen)

DakuTree: Hadn't even considered cases where the search is on the same line. Shall fix that.

ext/autocomplete/theme.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+class AutoCompleteTheme extends Themelet {
+	public function build_autocomplete(Page $page) {
+		$base_href = get_base_href();
+		// TODO: AJAX test and fallback.
+
+		$page->add_html_header("<script src='$base_href/ext/autocomplete/lib/jquery-ui.min.js' type='text/javascript'></script>");
+		$page->add_html_header("<script src='$base_href/ext/autocomplete/lib/tag-it.min.js' type='text/javascript'></script>");
+		$page->add_html_header('<link rel="stylesheet" type="text/css" href="http://ajax.googleapis.com/ajax/libs/jqueryui/1/themes/flick/jquery-ui.css">');
```
shish: `//server` rather than `http://server` will auto-detect the protocol, so that this can work on https as well as http sites :)

ext/autocomplete/style.css
DakuTree: Just noticed that the home ext isn't properly sorting HTML headers either, so the css isn't even properly being loaded there. Will fix that too..

ext/autocomplete/theme.php
```diff
@@ -0,0 +1,29 @@
...
+		$page->add_html_header("<link rel='stylesheet' type='text/css' href='$base_href/ext/autocomplete/lib/jquery.tagit.css' />");
+
+		$page->add_html_header("<script>
```
shish: should this be in `ext/autocomplete/script.js` so that it is included in the static bundle rather than embedded in each page?

DakuTree: Ah yeah it should be. Wrote the ext prior to making that change and this was to avoid it getting grabbed even if it wasn't loaded.

DakuTree:
Oh, one thing I forgot. You'll need to setup a GH_TOKEN variable in travis to stop composer throwing Github API errors (SEE: https://blog.wyrihaximus.net/2015/09/github-auth-token-on-travis/).
Supposedly they're in the process of getting rid of the need to do that, but it still seems required at the moment.

shish:
For autocomplete - add `ORDER BY tag`? Default order looks like random (in postgres at least)

(ORDER BY count would be nice if count was displayed somewhere, so that the user could see that they are getting the most popular tags first - but without seeing the count, I think that would be nearly as confusing as random)

ext/autocomplete/main.php
```diff
@@ -13,21 +13,28 @@ public function onPageRequest(PageRequestEvent $event) {
...
-			$res = $database->get_pairs("
-				SELECT tag, count
-				FROM tags
-				WHERE tag LIKE :search
-				AND count > 0
-				ORDER BY count DESC
-				$limitSQL", $SQLarr
-			);
+			$res = null;
+			$database->cache->get($cache_key);
```
shish: `$res = $database->cache->get($cache_key);`

(My bad I think? Was trying various combinations of caching rules >_>)

shish:
I'm getting autocomplete headers included before base headers for some reason o_O  Forcing them to be later with eg this fixes it:

```
               $page->add_html_header("<script src='$base_href/ext/autocomplete/lib/jquery-ui.min.js' type='text/javascript'></script>", 80);
               $page->add_html_header("<script src='$base_href/ext/autocomplete/lib/tag-it.min.js' type='text/javascript'></script>", 81);
               $page->add_html_header('<link rel="stylesheet" type="text/css" href="//ajax.googleapis.com/ajax/libs/jqueryui/1/themes/flick/jquery-ui.css">', 82);
               $page->add_html_header("<link rel='stylesheet' type='text/css' href='$base_href/ext/autocomplete/lib/jquery.tagit.css' />", 83);
```

DakuTree:
Shouldn't need to specify priority. That issue is caused by the html_headers not being sorted in some places (which is fixed in #556).

shish:
For case-insensitive autocomplete on postgres:
```
-                       $res = null;
                        $res = $database->cache->get($cache_key);
                        if(!$res) {
-                               $res = $database->get_pairs("
+                               $res = $database->get_pairs($database->scoreql_to_sql("
                                        SELECT tag, count
                                        FROM tags
-                                       WHERE tag LIKE :search
+                                       WHERE SCORE_STRNORM(tag) LIKE SCORE_STRNORM(:search)
                                        AND count > 0
                                        ORDER BY count DESC
-                                       $limitSQL", $SQLarr
+                                       $limitSQL"), $SQLarr
                                );
                                $database->cache->set($cache_key, $res, 600);
```

shish:
Everything looks good to me now \o/  I'm thinking to merge to develop now, and then new release next weekend if nothing shows up as totally broken during the week. Anything else you can think of to do before merging?

DakuTree:
The only thing I can think of is setting up the travis variable I mentioned above (if you haven't already). Other than that, everything seems good :smile:

shish:
Done /o/  And now to merge~

PR #556, opened 2016-06-18
fix and use get_all_html_headers()

shish:
this function was defined, but apparently not used ._.


DakuTree:
Should probably fix [ext/home/theme.php](https://github.com/shish/shimmie2/blob/master/ext/home/theme.php#L7) too.

PR #557, opened 2016-06-18
add composer.lock for scrutinizer-ci

shish:
Just seeing if this helps

https://scrutinizer-ci.com/g/shish/shimmie2/inspections/68bab58b-9038-4a01-b032-fcd8490cc657


DakuTree:
That seems to fix it. I'm guessing scrutinizer was failing due to being unable to properly grab the libs which use composer-asset-plugin.

Issue #547, opened 2016-05-22
multi-search with wildcard broken


shish:
image with 3 tags: foo1, foo2, bar

search for: foo*, baz

image shows up, I guess because number of tags matched = number of tags searched for

shish:
Gah, Daku fixed this for mysql, and then my attempt at merging the two search paths a little to stop the huge amount of deduplication broke it again, and it's 2am with work tomorrow morning >_<

We now have some much better test cases at least, will try and bring the fix back to all DB engines tomorrow /o/

DakuTree:
Looks like this broke again during some of the recent deduping commits.[¹](https://github.com/shish/shimmie2/commit/dc83d4ace72e55e2b70bf17f17d02cd5507f61ad#commitcomment-17922089) [²](https://github.com/shish/shimmie2/commit/bfa1dc20c01b50da775c55b93db9d16792b093d1#commitcomment-17922121)

Not sure how to go about fixing without reverting a bunch of the changes :|

shish:
Yeah, I think my re-fix-wildcards branch is closer to where we want to end up -- it keeps wildcards in their raw form (ie, "a string with a wildcard" rather than "an array of plain tags") until the last moment before executing SQL, so that we aren't losing that context.

I'm currently installing mysql on my laptop so that I can test that code path...

https://github.com/shish/shimmie2/blob/re-fix-wildcards/core/imageboard.pack.php#L985

PR #559, opened 2016-06-19
re-fix wildcards for mysql

shish:
see #547

The general approach here is to keep wildcards in their wildcard form until as late as possible, so that the searching function can treat them specially (as opposed to expanding them at the start and then having no idea which tags were requested and which are part of a wildcard expansion)

### DIFF
--- a/core/imageboard.pack.php
+++ b/core/imageboard.pack.php
@@ -316,7 +316,13 @@ class Image {
 		}
 
 		if(count($tags) === 0) {
-			$row = $database->get_row('SELECT images.* FROM images WHERE images.id '.$gtlt.' '.$this->id.' ORDER BY images.id '.$dir.' LIMIT 1');
+			$row = $database->get_row('
+				SELECT images.*
+				FROM images
+				WHERE images.id '.$gtlt.' '.$this->id.'
+				ORDER BY images.id '.$dir.'
+				LIMIT 1
+			');
 		}
 		else {
 			$tags[] = 'id'. $gtlt . $this->id;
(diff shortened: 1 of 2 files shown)


core/imageboard.pack.php
```diff
@@ -911,7 +928,7 @@ private static function build_accurate_search_querylet(
 				$database->scoreql_to_sql("
 					SELECT id
 					FROM tags
-					WHERE SCORE_STRNORM(tag) = SCORE_STRNORM(:tag)
+					WHERE SCORE_STRNORM(tag) LIKE SCORE_STRNORM(:tag)
```
shish: This is the actual fix

core/imageboard.pack.php
```diff
@@ -993,27 +1003,26 @@ private static function build_ugly_search_querylet(
...
 
-		$x = 0;
-		foreach($tag_search->variables as $tag) {
+		foreach($tag_querylets as $tq) {
 			$tag_ids = $database->get_col(
 				$database->scoreql_to_sql("
 					SELECT id
 					FROM tags
-					WHERE SCORE_STRNORM(tag) = SCORE_STRNORM(:tag)
+					WHERE SCORE_STRNORM(tag) LIKE SCORE_STRNORM(:tag)
```
shish: This is the actual fix

core/util.inc.php
```diff
@@ -412,6 +412,11 @@ function endsWith(/*string*/ $haystack, /*string*/ $needle) {
 	return (substr($haystack, $start) === $needle);
 }
 
+if(!function_exists("mb_strlen")) {  // D:
+	function mb_strlen($str) {return strlen($str);}
+	function mb_internal_encoding($enc) {}
+	function mb_strtolower($str) {return strtolower($str);}
+}
```
shish: Mostly for sake of simple unit testing

shish:
Merging since tests are happy /o/

PR #560, opened 2016-06-29
Added SetTagEvents to the mass tagger extension so the tags can be sanitized

sanmadjack:
The mass tagger wasn't updated when the tag sanitization code was moved to SetTagEvent, so things like aliases weren't working from the mass tagger.


shish:
Sorry for taking ages with this, I was looking and trying to figure out why this patch helps (it does help, but it shouldn't be necessary) -- ended up going over the whole codebase and making use of `Tag::explode()` consistent everywhere (ie, always explode tags on input, instead of scattering explosions throughout the codebase), so there should be much less flakiness now, and this should be fixed as a side effect ^^

PR #569, opened 2016-08-16
Fix bookmarklets

im-mi:
This should fix some of the issues with bookmarklets.

### Known limitations:
- Importing from Shimmie was and still is troublesome. It still has trouble with videos in this regard, from what I've tested. It also still does not handle ratings.
- If the source is using HTTPS, then the destination will have to as well. Otherwise it won't work in most popular browsers due to mixed content errors. It's recommended that the destination be using HTTPS for the highest compatibility.
- A few sites still don't work due to hotlink protection.


im-mi:
Oops, need to work on this a bit more

PR #574, opened 2016-08-29
Fix tags not being escaped for HTML in some cases (code injection vulnerability)

im-mi:
This fixes an issue where tags were not always escaped, allowing for code injection.

Try adding this tag to a post, then view it in the post list:
`'><script>alert("BOO!");</script>`


shish:
Cherry-picked into master too /o/

PR #576, opened 2016-09-01
Update pool description formatter (code injection vulnerability)

im-mi:
Fixes a code injection vulnerability where pool descriptions were allowed to have arbitrary HTML in them.

For reference, try using this as a pool description:
`<script>alert("BOO!");</script>`

It should be noted that I only guessed at this solution based on what the comments extension does. It seems to work, though.

### DIFF
--- a/ext/pools/theme.php
+++ b/ext/pools/theme.php
@@ -154,8 +154,9 @@ class PoolsTheme extends Themelet {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


shish:
That does look like a reasonable fix, individual text formatters aren't expected to *all* do HTML escaping /o/

PR #577, opened 2016-09-01
Fix "starts-with" header in tag list when escaping required

im-mi:
This fixes the "starts-with" headers* on the tags/alphabetic page. Before, the headers would be wrong if they started with an escaped character.

This also escapes the resulting header so that it no longer generates invalid HTML in such cases.

\* Note that these headers are only visible when paged tag lists is disabled.


shish:
Good catch, thanks <3

PR #578, opened 2016-09-02
Enhance zoom

im-mi:
# New features

- Added zoom support to webm, svg, and ico
    - Note that this makes webm no longer stretch to 100% width. An additional zoom mode could be added if the old behavior is desired.
- Re-apply zoom upon window resize
    - This affects the "Fit Height" and "Fit Both" zoom modes. Before, the user would have to refresh the page or toggle zoom settings to get the zoom to update.
- When loading an SVG, it now uses the size of the outermost SVG element. Before, the size could be reported incorrectly when there were nested SVG elements.

# Considerations

- The code that handles zooming is part of the handle_pixel extension. Making it handle other formats may be considered a breach of responsibility. Perhaps the zoom functionality could be moved to its own extension.
- This expands the usage of `.shm-main-image` and `#main_image` to other types of main post content. This has implications for CSS because now post content can be styled at once, regardless of whether it is pixel, webm, svg, or ico. Personally, I feel that this is a good change.

# Limitations

- I didn't expand support to flash-based video. I can't test it (flash video doesn't work for me), and its height is hard-coded to 480px which makes me feel that changing it may break something.
- Click-to-zoom is disabled for webm so that the player controls are still accessible.
- Click-to-zoom is disabled for SVG because said format can contain interactive content and I couldn't get SVG click detection to work in any browser anyways.

# Issues

- Supporting zooming means that the content's width and height can't be set directly in the HTML. This may cause jitteriness because then determining the content's size is deferred until handle_pixel's zoom script runs. Note that this issue already affected pixel images.
- The SVG theme uses a double-elemented method of embedding SVGs (`<object><embed/><object>`). This could potentially cause issues in some browsers because it's currently set such that the zoom code only interacts with the outer `<object>` element. Regardless, it works on the latest desktop Chrome, Edge, and Firefox.

### DIFF
--- a/ext/handle_ico/theme.php
+++ b/ext/handle_ico/theme.php
@@ -4,7 +4,8 @@ class IcoFileHandlerTheme extends Themelet {
 	public function display_image(Page $page, Image $image) {
 		$ilink = make_link("get_ico/{$image->id}/{$image->id}.ico");
 		$html = "
-			<img id='main_image' src='$ilink'>
+			<img id='main_image' class='shm-main-image' alt='main image' src='$ilink'
+			data-width='{$image->width}' data-height='{$image->height}'>
 		";
 		$page->add_block(new Block("Image", $html, "main", 10));
 	}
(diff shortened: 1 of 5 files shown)


im-mi:
Updated with functioning (?) SVG support

im-mi:
Nothing left to add here.

shish:
All looks good /o/  Sorry for taking forever to review this >_<

PR #580, opened 2016-09-02
Mass tagger fixes

im-mi:
- Clear selected images on load in case they were autocompleted by the browser.
    - Otherwise you could end up with images that were selected but didn't have the selection effect.
- Use CSS for mass tagger button styling
    - Uses the class name `mass-tagger-selected`
    - Makes it easier to style the selection effect


shish:
All looks good :)

PR #581, opened 2016-09-03
Store tag lists logically inside HTML lists

im-mi:
**Update:** This PR has been superseded by https://github.com/shish/shimmie2/pull/583

This change wraps the sidebar tag lists into HTML lists. Before, they were structured as a bunch of loose elements punctuated with `<br/>` tags. This makes the resulting HTML a little more organized and makes it much easier to style the tag lists.

The resulting lists were given the class `tag_list`, similar to the other classes used by the `tag_list` extension including `tag_name`, `tag_info_link`, and `tag_count`.

Minimal changes were made to keep the built-in themes looking the same. However, the Material theme is very slightly different now since it uses a 1px larger font for `<ul>`. This I didn't feel was worth changing because that theme's style is very much set in stone.

### DIFF
--- a/ext/tag_list/style.css
+++ b/ext/tag_list/style.css
@@ -1,0 +1,5 @@
+.tag_list {
+	list-style-type: none;
+	margin: 0;
+	padding: 0;
+}
--- a/ext/tag_list/theme.php
+++ b/ext/tag_list/theme.php
@@ -61,14 +61,18 @@ class TagListTheme extends Themelet {
 			$category = $split[0];
 			$tag_html = $split[1];
 			if(!isset($tag_categories_html[$category])) {
-				$tag_categories_html[$category] = '';
+				$tag_categories_html[$category] = '<ul class="tag_list">';
 			}
-			$tag_categories_html[$category] .= $tag_html . '<br />';
+			$tag_categories_html[$category] .= "<li>$tag_html</li>";
 
 			if(!isset($tag_categories_count[$category])) {
 				$tag_categories_count[$category] = 0;
 			}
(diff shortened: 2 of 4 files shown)


im-mi:
... or maybe it would be better represented as a table?

im-mi:
After some though, I think that tables make more sense for the kind of data being displayed. And, after some experimentation, I've found that it's possible to keep the same general look while doing so.

## Default tag list
The tag list's original look, using the default theme.
default

## Default tag list using tables
And here's how it looks after tabled tag lists are introduced. No additional styles have been applied yet.
default as table

## Default tag list using tables + styling
Applying a style can make the tag list look like much like it did before, with one caveat (see below).
```
TABLE.tag_list {
	border-collapse: collapse;
}
TABLE.tag_list>THEAD>TR {
	display: none;
}
TABLE.tag_list>TBODY>TR>TD {
	display: inline;
}
```

And the result:

default as table style

### Caveats

In some browsers, this results in there being slightly more padding between the parts of the tag list item. By comparing with the unmodified look you can see that Firefox adds a little bit of extra padding. Other browsers like Chrome display exactly as before.

## Custom theme example
Here I use a custom theme to demonstrate the structured look that tables can provide to tag lists.

custom with tables style

### Remarks
Additionally, with tables it could be made possible to re-sort tag lists by clicking column headers.

## Other considerations
This change would notably require updating any existing themes if they were to keep the same look. This might be reason enough to hold off on this change. Though, either way I'll be using tables for my tag lists, if not through an official branch then through a custom tag_list theme.

dali99:
This looks great +1!

PR #583, opened 2016-09-06
Use tables for tag lists

im-mi:
# Overview

This change organizes the sidebar tag lists into tables. Originally I had used lists, but then later decided that tables were a better fit (see https://github.com/shish/shimmie2/pull/581).

Additionally, this updates all of the built-in themes to keep the same general look.

# Rationale

Tag lists are currently structured as a bunch of loose elements punctuated with `<br/>` tags. Organizing them as a table has several advantages:
- Cleaner HTML
- Easier to style
- Sortable*

\* This feature is implemented and works, but it requires that the table headers be visible. All of the built-in themes hide the table headers.

# Issues

- The spacing is slightly different now. Often, the tag list is shifted downward by a pixel or two. This also varies a little across browsers, with Edge notably increasing the line height sometimes by a pixel or so. The differences here are negligible, however.
- Copying from the tag list in Edge now produces three spaces between columns instead of one like it did before.
- This change will likely affect any non-standard themes. However, getting the original look back should be easy in most cases (a few lines of CSS).

# Other Notes

Also added were various classes to aid in styling:
- The tag list itself has the `tag_list` class.
- The columns and cells have classes that match their contents (ex. `tag_name`, `tag_name_cell`, and `tag_name_column`)

### DIFF
--- a/ext/tag_list/theme.php
+++ b/ext/tag_list/theme.php
@@ -229,13 +229,13 @@ class TagListTheme extends Themelet {
 		// if($n++) $display_html .= "\n<br/>";
 		if(!is_null($config->get_string('info_link'))) {
 			$link = html_escape(str_replace('$tag', $tag, $config->get_string('info_link')));
-			$display_html .= ' <td class="tag_info_link_cell"><a class="tag_info_link'.$tag_category_css.'" '.$tag_category_style.'href="'.$link.'">?</a></td>';
+			$display_html .= '<td class="tag_info_link_cell"> <a class="tag_info_link'.$tag_category_css.'" '.$tag_category_style.'href="'.$link.'">?</a></td>';
 		}
 		$link = $this->tag_link($row['tag']);
-		$display_html .= ' <td class="tag_name_cell"><a class="tag_name'.$tag_category_css.'" '.$tag_category_style.'href="'.$link.'">'.$h_tag_no_underscores.'</a></td>';
+		$display_html .= '<td class="tag_name_cell"> <a class="tag_name'.$tag_category_css.'" '.$tag_category_style.'href="'.$link.'">'.$h_tag_no_underscores.'</a></td>';
 
 		if($config->get_bool("tag_list_numbers")) {
-			$display_html .= " <td class='tag_count_cell'><span class='tag_count'>$count</span></td>";
+			$display_html .= "<td class='tag_count_cell'> <span class='tag_count'>$count</span></td>";
 		}
 
 		return array($category, $display_html);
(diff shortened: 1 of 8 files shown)


im-mi:
Copied from https://github.com/shish/shimmie2/pull/581#issuecomment-244641115

---

I've decided that using tables is a better choice for tag lists. And, after some experimentation, I've found that it's possible to keep the same general look while doing so.

## The original look
This is the tag list's original look, using the default theme.

default

## The tabled look
And here's how it looks after tabled tag lists are introduced. No additional styles have been applied yet.

default as table

## Getting back the original look
Getting back the original look is now a matter of applying the right style. It's possible to do so, but with one caveat (see below). The style I will use is as follows:
```css
TABLE.tag_list {
	border-collapse: collapse;
}
TABLE.tag_list>THEAD {
	display: none;
}
TABLE.tag_list>TBODY>TR>TD {
	display: inline;
}
```

And the result:

default as table style

### Caveats

Some browsers add a small amount of extra padding. By comparing with the original look you can see that Firefox is one of them. Others, like Chrome, display exactly as before.

## Custom theme example
Here I use a custom theme to demonstrate the structured look that tables can provide to tag lists.

custom with tables style

## Other considerations
This change would require updating existing themes if they were to keep the same look. This might be reason enough to hold off on this change. Though either way, I'll be using tables for my tag lists (if not through an official branch, then through a custom tag_list theme).

It does have its benefits, though, and would make it possible to do other tabley things like re-sorting when clicking column headers.

im-mi:
In this update:

### Change 1: Made the spacing more like it was before

It's still a few pixels off in some cases, but the difference is now negligible (a pixel or two difference).



### Change 2: Fix clipboard spacing

This brings back the original behavior of having spaces between the columns when the tag list gets copied to the clipboard (tested in Chrome, Edge, and Firefox).

One caveat: Edge now adds _three_ spaces instead of one.

im-mi:
I have nothing left to add to this, though there are a few potential issues with merging it:
- It's untested in older browsers.
- The CSS is a bit more complicated as a result. Displaying a table as inline might feel a little hacky.
- It isn't backward compatible. Most non-standard themes would need to be updated.
- It might not be the direction you want to take things.

shish:
It is looking pretty good, sorry I've not got round to looking in depth yet, I've been on a pretty hectic business trip for the last couple of weeks, and for next week too - I'm back at home in a week and a bit, and also this weekend is looking fairly empty so I will attempt to catch up with my open source projects then ^^

PR #586, opened 2016-09-12
Image info tweaks

im-mi:
- Hide the rating editor outside of edit mode
  - Instead, it shows the human-friendly name of the current rating. Note that this has no effect when JS is disabled because then there's no choice but to show the editor.
- Show image rating in image info box even when not logged in
  - I found it odd that the image's rating would disappear if you weren't logged in. This fixes that.
- Removed trailing period from "Parent: None." for consistency
  - No other image editor properties that I know of use a trailing period

### DIFF
--- a/ext/rating/main.php
+++ b/ext/rating/main.php
@@ -103,9 +103,7 @@ class Ratings extends Extension {
 	}
 	
 	public function onImageInfoBoxBuilding(ImageInfoBoxBuildingEvent $event) {
-		if($this->can_rate()) {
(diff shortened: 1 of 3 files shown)


im-mi:
New in this update:
- Only show image rater if user can rate
  - If the user can't rate, then it shows just the rating instead of the editor. This behavior is consistent with the other image info editors.

im-mi:
I have nothing left to add to this. Anything amiss? Otherwise it should be an easy merge.

PR #587, opened 2016-09-14
Fix comment-delete injection

im-mi:
# Overview

This fixes a JS code injection vulnerability that affected the "delete comment" dialog. By using a crafted comment, it was possible to inject a small amount of Javascript (~50 characters) to be executed whenever a user tries to delete an affected comment.

# Examples

Below are examples of crafted comments and the effects they would have without this fix.

### Show a message
```
')&&alert('BOO!
```

### Make a comment undeletable (simple)
```
')==('
```

### Make a comment undeletable (advanced)
```
hitler did nothn wrong[align=right]')==('[/align]
```

# Notes

In order to correctly escape the strings used in the JS-based comment deletion prompt, it was necessary to unescape the result of the TextFormattingEvent because it had already been escaped for HTML. While doing so felt a little dirty, there was no other way besides making deeper changes.

### DIFF
--- a/core/util.inc.php
+++ b/core/util.inc.php
@@ -16,6 +16,16 @@ function html_escape($input) {
 }
 
 /**
+ * Unescape data that was made safe for printing into HTML
+ *
+ * @param $input
+ * @return string
+ */
+function html_unescape($input) {
+	return html_entity_decode($input, ENT_QUOTES, "UTF-8");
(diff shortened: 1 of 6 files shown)


shish:
Gah D: I wish XHP would hurry up and be a thing so string concatenation can die ~_~

PR #600, opened 2017-03-10
Bump PHP version up to 5.6 & Linting/fixes

jgen:
PHP 5.5 has been EOL for ~7 month now. (http://php.net/eol.php and http://php.net/supported-versions.php)
This change bumps up the version of PHP in the Read-me, and used by Travis for testing to version 5.6.
It adds phpunit as an explicit dependency to Composer in order to get the Travis builds working on PHP 7/7.1
Also includes a bunch of changes for linting, PHPDoc comments for type hints, and some changes suggested by Scrutinizer-CI.


composer.json
```diff
@@ -19,7 +19,8 @@
 	],
 
 	"require" : {
-		"php" : ">=5.4.8",
+		"php" : ">=5.6",
+		"phpunit/phpunit" : "5.*",
```
DakuTree: This should probably be put this under "require-dev" as to avoid including it in production.

jgen: Good point! Thanks @DakuTree :+1:

shish:
All looking pretty good; supporting legacy PHP is one of the things that makes me sad whenever doing PHP work, so super-happy to get a version bump :D

PR #601, opened 2017-03-10
Updated/removed various libs + tweaks

DakuTree:
Thought it would be good to get this out before release.

Main things in the PR:
* Moved from composer-asset-plugin to asset-packagist.
  * This means no more need for a globally installed plugin in development.
  * Works exactly the same. More info @ https://asset-packagist.org/
* Removed getID3 and replaced it with mime_content_type() & jsmediatags where necessary.
* handle_mp3 now uses the html5 audio element rather than a flash player.
* Updated PHP-Akismet so it works without warnings on PHP7+
* Removed password_compat as it's no longer needed as we aren't supporting below PHP5.5.


jgen:
Thanks Daku!

shish:
Excellent :D

PR #604, opened 2017-03-13
Some more linting

jgen:
Mostly just a little bit more linting.
Also, remove the wrapper for mb_strlen and mb_internal_encoding, as they have been in core PHP since version 4.0.6. (mb_strtolower has also been in core PHP since version 4.3.0).


shish:
All looking good, thanks :)

PR #605, opened 2017-03-18
Fix SVG XSS

DakuTree:
SVG allows inline Javascript execution when fetched as an \<object\>. Using \ stops this.  
Example SVG using alert(): [example.svg](https://codeanimu.net/random/archive/uploads/example.svg)

Some related reading:
* https://bartblaze.blogspot.co.uk/2016/11/nemucod-downloader-spreading-via.html
* https://security.stackexchange.com/q/143141
* https://stackoverflow.com/questions/33387722
* https://www.w3.org/wiki/SVG_Security

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -25,12 +25,6 @@
 	RewriteRule ^(.*)$ index.php?q=$1&%{QUERY_STRING} [L]
 </IfModule>
 
-<IfModule mod_php5.c>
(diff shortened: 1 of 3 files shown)


jgen:
Normally I'd wait for @shish to merge it in, but given the security nature of the issue I'm going to go ahead and merge it into the develop branch.

Issue #599, opened 2017-03-08
Release time?


jgen:
We should probably push out an official release...
Its been ~2 years since the last one, and there has been lots of change in the meantime. (Composer, new theme from @dali99, many bug fixes, etc).

Though this means packaging up the composer dependencies into a zip for consumption.
I took a brief look into it, but I'm not sure how best to go about it. (It would be great if someone can document on the wiki how if/when the make an attempt.)

DakuTree:
Release with composer should just be a simple ```composer install --no-dev```.  
Maybe also clear up the files that aren't needed for production (.git, eclipse, composer, scrutinizer & travis files for example).

Should probably try and grab the stray commits that also managed to get committed to master by mistake, and pull them over to master. Few odd bug fixes in there.

jgen:
Thanks Daku (and thanks for your work on moving over to composer!).
Those are some good ideas for packaging up a release. I didn't realize it would be that simple -- various other docs mention performance tweaks you should make to composer in production.
(Though in retrospect I don't think we are using any of the auto-loading features of composer, so perhaps that isn't relevant).

Good point about stray commits on master. Are you saying we should merge them into the develop branch? (merge from master to master seems odd?)

Thanks again!

DakuTree:
> (Though in retrospect I don't think we are using any of the auto-loading features of composer, so perhaps that isn't relevant).

We are using the composer autoloader at the moment as it's a bit easier to work with: https://github.com/shish/shimmie2/blob/develop/core/_bootstrap.inc.php#L12  
Extension specific libs are still loaded manually though, as to avoid loading them when the extension is disabled.

> ... various other docs mention performance tweaks you should make to composer in production.

Didn't know there was any performance tweaks for composer in production..  
Having a quick look, it seems like ```--optimize-autoloader``` would help us here. [This](http://mouf-php.com/optimizing-composer-autoloader-performance) post reports a 37%~ speed increase.

> Good point about stray commits on master. Are you saying we should merge them into the develop branch? (merge from master to master seems odd?)

Yeah, meant to say pull them over to develop.

jgen:
Given the amount of time between releases, all of the various changes/tweaks, bug fixes, using composer, using phpunit now, etc. etc. -- I'm wondering if perhaps we might consider bumping the version number up from 2.5.5+ to something like 2.6.0 before releasing?

jgen:
[...]

PR #606, opened 2017-03-23
Bump the version number before releasing

jgen:
Given the amount of time (and churn) between releases, we should probably bump the version number before releasing (#599). 
(Note: In the official release we should remove the "+" from the version number.)


shish:
FWIW my version-bumping process so far has been "commit 'version X' to master, tag that commit, merge master into develop, add a commit to develop setting the version to 'version X+'" (so that when viewed linearly, the commit history is X (master), X+ (develop), Y (master), Y+ (develop), etc)

Though as long as master ends up tagged and released with a sensible number I don't think it makes that much difference :P

jgen:
Thanks @shish that's a good idea. I'll cancel this pull request and wait until we merge from develop into master before bumper the version number.

PR #608, opened 2017-03-25
Merge from develop into master

jgen:
It's happening~!


shish:
Seems about time, and tests look good :)

PR #611, opened 2017-04-25
Fix error upon bulk add.

thomas-hori:
Fixes "PHP Fatal error:  Uncaught TypeError: Argument 2 passed to
TagSetEvent::__construct() must be of the type array, string given"
upon bulk add.


thomas-hori:
Thanks, have done.

jgen:
Thank you! :+1:

Issue #612, opened 2017-05-11
Archive Handler chokes on upload. (Current Master and Dev branches)


greysondn:
Edit: See the update below. The following bit is kept as a matter of record, and the apparent fix - although in staging - still needs merged.

---

There's a bit of logs from the Freenode IRC, for starters.

```
[15:45] == greysondn [b80f3f36@gateway/web/freenode/ip.184.15.63.54] has joined #shimmie
[15:46] <greysondn> lo room. anyone got a few minutes to try to tackle an issue I'm having with zip archive uploads? Will post log in a moment.
[15:48] <greysondn> Log entry: https://pastebin.com/fpxTwM50
[15:48] <greysondn> That was my sanity test zip after updating my server. This has happened with multiple zip files.
[16:04] <greysondn> I've grabbed the latest dev branch version, will test momentarily. I think I was on stable, can't swear to it. Also can't be here all day so I may drop out before anyone get to it.
[16:27] <greysondn> can confirm problem in dev.
[17:04] <greysondn> Yeah, it's pretty borked. I'mma give up trying
[17:04] <greysondn> A related problem was found in the mass uploader; most likely the tags aren't being properly expanded for the tagger.
[17:04] <greysondn> https://github.com/shish/shimmie2/issues/575
[17:05] <greysondn> I'll file an issue backtracking to this bit of chat "soon".
```

As noted, this seems - at least on the surface - to have similarities with #575 . The tag manager doesn't like the tags. I don't speak PHP and don't swear to know, but if I had to bet, the fix exists and only needs integrated properly.

Data about my environment can be made available upon request; however, I am not online often these days. I host Shimmie locally to manage my personal porn stash. While I'm here, great work! Do keep it up!

---

Update:

pull #611 seems to fix the first part of the problems with the archive uploader. In Master, a new issue develops: https://pastebin.com/tw2hfzPi

I am preparing to test Dev. Many thanks to chat's Frogging101 (not sure of dev handle on github) for help this far.

greysondn:
Can confirm same behavior from updated section in Dev branch, toying around a bit.

greysondn:
Fix tested and working in dev branch.

File pointer:
https://github.com/shish/shimmie2/blob/develop/ext/handle_archive/main.php#L37

Delete that conditional and everything behaves just fine. This file evidently has no theme, which causes the second part of this issue.

[...]

PR #610, opened 2017-04-23
bulk_add_csv: Run Tag::explode() on the tags

Frogging101:
Fixes #575

Issue #575, opened 2016-08-31
Bulk upload doesn't work


sanmadjack:
Get this when I try to do a bulk add:

```
<br />
<b>Warning</b>:  assert(): Assertion &quot;is_array($metadata[&quot;tags&quot;])&quot; failed in         <b>/octank/srv/shimmie_rand/ext/upload/main.php</b> on line <b>35</b><br />
<br />
<b>Warning</b>:  Invalid argument supplied for foreach() in     <b>/octank/srv/shimmie_rand/ext/not_a_tag/main.php</b> on line <b>38</b><br />
<br />
<b>Warning</b>:  array_merge(): Argument #1 is not an array in     <b>/octank/srv/shimmie_rand/ext/image/main.php</b> on line <b>334</b><br />
<b>Catchable fatal error</b>:  Argument 2 passed to TagSetEvent::__construct() must be of the type     array, null given, called in /octank/srv/shimmie_rand/ext/image/main.php on line 339 and defined in     <b>/octank/srv/shimmie_rand/ext/tag_edit/main.php</b> on line <b>105</b><br />
```

I poked into it, and it looks like the list of tags from the file name aren't being exploded, so it tries to array_merge a string of the new tags and an array of the existing tags. This causes it to fail on the tag set event.

sanmadjack:
I'm on the latest develop branch commit as of right now.

PR #614, opened 2017-05-14
Ensure that metadata['tags'] is always an array

jgen:
This pull-request is somewhat related to the following issues/bug/pull-requests:
#575,  #610, #611, #612 

We should be more consistent and should ensure that `metadata['tags']` is always an array, in order to prevent issues like the previously mentioned bugs.

### DIFF
--- a/ext/cron_uploader/main.php
+++ b/ext/cron_uploader/main.php
@@ -304,6 +304,10 @@ class CronUploader extends Extension {
 
 	/**
 	 * Generate the necessary DataUploadEvent for a given image and tags.
+	 *
+	 * @param string $tmpname
+	 * @param string $filename
+	 * @param string $tags
 	 */
 	private function add_image($tmpname, $filename, $tags) {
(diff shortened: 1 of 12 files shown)


ext/cron_uploader/main.php
```diff
@@ -315,7 +319,7 @@ private function add_image($tmpname, $filename, $tags) {
 		$metadata = array();
 		$metadata ['filename'] = $pathinfo ['basename'];
 		$metadata ['extension'] = $pathinfo ['extension'];
-		$metadata ['tags'] = ""; // = $tags; doesn't work when not logged in here
+		$metadata ['tags'] = array(""); // = $tags; doesn't work when not logged in here
```
shish: should this be an empty array, rather than an array with one empty string?

jgen: Hmm..good point -- thanks @shish! :)

shish:
This all looks very sensible \o/

PR #643, opened 2018-08-14
Add missing escaping of ffmpeg shell command during video thumbnail generation

ThePadawan:
When generating video thumbnails, the exec() call to ffmpeg is currently only escaped in the case where the "ignore video scale" configuration is set to true. In the "else" branch, the raw command is executed (!).


shish:
AFAIK we're only dealing with trusted inputs there, but escaping those inputs should be a no-op, and it'd catch bad data if it did somehow get there, so merging :)

ThePadawan:
Thanks! I actually ran into some issues when using ffmpeg on Windows which resulted in an error like "Executable could not be found" when using escapeshellcmd(), but I'm investigating this separately from this PR.

PR #653, opened 2019-06-01
 Fixes for various issues

sanmadjack:
Fixed an issue when null queries were passed to some themes. Just needed the parameter specs changed to allow nulls.

### DIFF
--- a/core/config.php
+++ b/core/config.php
@@ -113,7 +113,7 @@ abstract class BaseConfig implements Config
 {
     public $values = [];
 
-    public function set_int(string $name, ?int $value): void
+    public function set_int(string $name, ?string $value): void
     {
         $this->values[$name] = parse_shorthand_int($value);
         $this->save($name);
--- a/core/config.php
+++ b/core/config.php
@@ -18,7 +18,7 @@ interface Config
     /**
      * Set a configuration option to a new value, regardless of what the value is at the moment.
      */
-    public function set_int(string $name, ?int $value): void;
+    public function set_int(string $name, ?string $value): void;
 
     /**
      * Set a configuration option to a new value, regardless of what the value is at the moment.
--- a/core/imageboard/image.php
+++ b/core/imageboard/image.php
@@ -570,6 +570,8 @@ class Image
     {
         global $database;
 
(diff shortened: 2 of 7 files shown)


sanmadjack:
Added a commit that corrects an issue preventing the cron upload from working at all. Then check for if the key wasn't generated never succeeded, so it would never generate the key, which prevents the whole thing from working.

sanmadjack:
Added another commit, to take care of an issue when saving settings in shorthand, like 1M, Since set_int was changed to only accept int it was stripping out the letters, causing settings on the config page to lose that information.

sanmadjack:
Added array_unique to set_tags to prevent primary key violations when upload conflict is set to merge

sanmadjack:
Another commit, this one to take care of an issue with my previous change to prevent cron upload warnings. Since processed items were removed from the queue, comparing the queue length to $i resulted in some items being skipped. Changed the test to just if the length of the queue is greater than 0. Also now using array_pop to pull from the queue (array_shift has an unnecessary performance hit, since the import queue is randomized anyway, and it's imho cleaner than doing it at the end with unset).

sanmadjack:
The changes to cron upload to use RecursiveDirectoryIterator happened to break path-based tag handling, since it removed $subdir from the equation. Since I'm already working on my pull request for merging tag handling, I've gone ahead and just changed cron upload to use the path_to_tags like everything else does, and resolved the issue the lack of $subdir introduced.

sanmadjack:
After switching to develop I found that the new tagging event requires at least a tag or it errors. I didn't want to take the leap of changing some of the image creation logic, so I tested adding the tags immediately upon image creation, rather than right after. This line is from the code that handled image creation:
        $metadata ['tags'] = []; // = $tags; doesn't work when not logged in here
I tried changing it to:
        $metadata ['tags'] = Tag::explode($tags); 
And removing the later tagging step and it seems to be working okay in my tests. I don't know why it was originally commented out like that, if you think it's for a good reason let me know and I'll remove it from this pull request.

shish:
All looks good to me, thanks :D

PR #654, opened 2019-06-02
Cron upload changes

sanmadjack:
Changes path-to-tag to merge path tags and filename tags.
Changes cron upload to mirror the subdir structure on output.
Changes the path-to-tag regexp to allow extensions with numbers.
Added early escape to the cron upload process when a transaction-invalidating error occurs.

### DIFF
--- a/ext/cron_uploader/main.php
+++ b/ext/cron_uploader/main.php
@@ -273,6 +273,11 @@ class CronUploader extends Extension
                 $this->move_uploaded($img[0], $img[1], false);
             } catch (Exception $e) {
                 $this->move_uploaded($img[0], $img[1], true);
+                if (strpos($e->getMessage(), 'SQLSTATE') !== false) {
+                    // Postgres invalidates the transaction if there is an SQL error, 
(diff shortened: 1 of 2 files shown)


ext/cron_uploader/main.php
```diff
@@ -273,6 +273,11 @@ public function process_upload(int $upload_count = 0): bool
                 $this->move_uploaded($img[0], $img[1], false);
             } catch (Exception $e) {
                 $this->move_uploaded($img[0], $img[1], true);
+                if (strpos($e->getMessage(), 'SQLSTATE') !== false) {
```
shish: I wonder what SQL errors can happen? Is there some way to avoid hitting them in the first place?

sanmadjack: Basically just bugs. The only error I encountered was a primary key violation caused by the item merge code sending duplicate tags to be written. The array unique change I submitted in the other pull request resolved most of those, but I found some other ways to trigger it that I'm going to make a ticket to discuss in.

PR #659, opened 2019-06-06
Bulk action extension and friends

sanmadjack:
Alright, this adds a new "Bulk Action" block to the left side of the post list. The extension provides a BulkActionBlockBuildingEvent that other extensions can use to add actions to the bulk action block. Upon the submission of a bulk action the extension sends a BulkActionEvent that other extensions can catch in order to perform the work for the bulk action. The extension handles gathering the items to perform the bulk action on, based on either the currently active query or the manually selected items, and passes them in the BulkActionEvent. If the number of items from a query are more than 100 it sends them in batches of 100.

The base extension includes actions or deleting, tagging, and setting the source of images. Regen thumbnails and ratings extensions have been modified to work with this extension. Other core extensions have been modified to not add their bulk controls in favor of the new ones provided here.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -27,7 +27,7 @@
 
 <IfModule mod_expires.c>
 	ExpiresActive On
-	<FilesMatch "([0-9a-f]{32}|\.(gif|jpe?g|png|css|js))$">
+	<FilesMatch "([0-9a-f]{32}|\.(gif|jpe?g|png|webp|css|js))$">
 		<IfModule mod_headers.c>
 			Header set Cache-Control "public, max-age=2629743"
 		</IfModule>
@@ -46,6 +46,7 @@
 AddType image/jpeg  jpg jpeg
 AddType image/gif   gif
 AddType image/png   png
+AddType image/webp  webp
 
 #EXT: handle_ico
 AddType image/x-icon   ico ani cur
--- a/README.markdown
+++ b/README.markdown
@@ -86,7 +86,7 @@ different enough to be a pain.
 Various aspects of Shimmie can be configured to suit your site specific needs
 via the file `data/config/shimmie.conf.php` (created after installation).
 
-Take a look at `core/sys_config.inc.php` for the available options that can
+Take a look at `core/sys_config.php` for the available options that can
 be used.
 
 
@@ -116,7 +116,7 @@ new UserClass("moderator", "user", array(
 ));
 ```
 
-For a list of permissions, see `core/userclass.class.php`
+For a list of permissions, see `core/userclass.php`
 
 
 # Development Info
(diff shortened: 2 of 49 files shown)


sanmadjack:
A few additional items. Bug fixes:

1. Changed to prevent writing duplicate image tag IDs
2. Changed cron upload so that an unrecognized file type results in an error instead of a success
3. Bug fixes and consolidation of various thumbnail and resize functionalities

Changes:

1. Changed resize/rotate extensions to use replace image event
2. Added content-disposition header to image responses to provide a human-friendly filename when saving
3. Added more bulk thumbnail regeneration tools
4. Tweaks to bulk actions to correct totals when batching items
5. Added thumb_scaling option for generating high-dpi thumbnails

sanmadjack:
A few more bug fixes:

1. Changed clamp function to allow null values
2. Moved ImageResizeException to the core space so that the core space image resize code can use it
3. Updated an install step in the rating extension to be pgsql compatible
4. readme corrections

Changes:

1. Added <label> to extension name in extension manager so the name can also be clicked to enable/disable an extension
2. Added OnTagTermParse to rating extension
3. Updated copyright notices to 2019
4. Added better error handling for GD code, checking all the things for returning false. GD code failing for any reason during a replace action could result in both image files being lost, this takes care of that.
5. Changes to bulk actions, passing full ID arrays instead of chunked image arrays. Some extension ideas I had were severely limited by chunking the items, so I changed it to be all the matching item IDs at once. This also means no juggling completed counts to keep them accurate.
6. Changed the bulk actions to have a separate identifier from the button name. Should have done this at first.
7. Added transaction handling to cron upload so that subsequent errors don't result in images that have already moved to the uploaded folder from being wiped from the database.
8. Changed cron upload output folders to use subfolders based on the timestamp of the current run. This is to prevent writing over files in the error folder that happen to have the same name and path, effectively losing the data.
9. Added additional error and information logging to cron upload, and a final count of imported/merged/failed.

New features:
1. Added option to detect file type based on header bytes
2. Added transcode extension to allow admins to convert images to other types (for instance, converting PNG to more efficient lossless webps, not that I made this just so I could do that). It also allows uploading image formats that aren't compatible with the web, such as TIFF and PSD, by automatically transcoding them to a supported file format.

arthurmv:
Great! When we could enjoy all upgrades?

shish:
There's a bunch of code so I'mma try it out on my testbed and see it working in practice :)

It's about time for a new release, so I'll do that once the develop branch settles a bit

sanmadjack:
Just checked in some fixes, make sure you test with those too

shish:
For ease of code review I'm thinking I might cherry-pick some individual commits for self-contained fixes (eg "Changed clamp function to allow null values"), which should make the new-feature changes easier to check ^^

sanmadjack:
Sounds perfectly reasonable, I kind of three a lot at you all at once here

shish:
Merged manually :3

PR #663, opened 2019-06-18
Bugfixes and small changes

sanmadjack:


core/imageboard/misc.php
```diff
@@ -24,7 +25,8 @@ function move_upload_to_archive(DataUploadEvent $event): void
 /**
  * Add a directory full of images
  *
- * #return string[]
```
shish: In these cases I've been using PHP's native type hints instead of duplicating the same info in the comment, as duplication doesn't seem to add anything - except in the case of arrays, where PHP doesn't support typed arrays, but PHPDoc comments do

sanmadjack: Thanks for the feedback, I'll keep that in mind

PR #667, opened 2019-06-21
Media extension

sanmadjack:
With a lot of the changes I've made to allow thumbnailing for more formats and centralizing more imagemagick, GD, and ffmpeg functionality it didn't seem right that some settings required for them to work were from extensions that the user may or may not have enabled. So, I set up this extension and migrated the common ffmpeg and convert path settings into it. I set it as a default extension, so it's always enabled. All of the other extensions now look to this extension's settings for those values, and it will auto-migrate them from the old locations. 

I've also migrated the shared media functions (like image resizing and ffmpeg calls) into the extension as static functions and triggerable events. The ones that are still static functions will likely become events later, I want to see how some other work I'm doing turns out to see what the best implementation would be. I've also moved the engine/format support constant arrays to here from the transcode extension, since it makes since for them to be in a central location, and allows other extensions to have a supported file reference that can change dynamically based on the selected media engine.

### DIFF
--- a/core/basethemelet.php
+++ b/core/basethemelet.php
@@ -58,7 +58,7 @@ class BaseThemelet
             $tsize = get_thumbnail_size($image->width, $image->height);
         } else {
             //Use max thumbnail size if using thumbless filetype
-            $tsize = get_thumbnail_size($config->get_int('thumb_width'), $config->get_int('thumb_height'));
+            $tsize = get_thumbnail_size($config->get_int(ImageConfig::THUMB_WIDTH), $config->get_int(ImageConfig::THUMB_WIDTH));
         }
 
         $custom_classes = "";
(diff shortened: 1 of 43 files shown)


shish:
(Sorry for the delay in reviewing this - I'm currently doing two 24/7 oncall shifts in parallel, during performance-review season, and then spending all my time outside of work moving house x__x)

sanmadjack:
> (Sorry for the delay in reviewing this - I'm currently doing two 24/7 oncall shifts in parallel, during performance-review season, and then spending all my time outside of work moving house x__x)

Sounds too familiar. Good luck, no hurry.

sanmadjack:
Rebased this pr

shish:
This should be merged now, minus the user_config stuff that's in the other pull request

PR #675, opened 2019-07-07
Remove build_ugly_search_querylet

shish:
now that the accurate search doesn't use subqueries, perhaps mysql can accept it?


sanmadjack:
I wasn't able to do any mysql testing, were you able to?

Issue #525, opened 2015-10-14
ext_manager page should show when extensions can't be enabled


DanielOaks:
The `ext_manager` should show when extensions aren't able to be enabled due to `$db_support` not supporting the currently used db.

I was trying to work out why a certain extension wasn't showing at all, even though it was enabled on the `ext_manager` page and everything looked good. Maybe putting a little orange mark or something next to the extension name saying something like "This won't be enabled due to the currently selected database. This extension can be used on the mysql, sqlite databases" when you hover over it?

dali99:
:+1:

shish:
This got done a while ago

PR #680, opened 2019-08-01
Proposed extension info change to allow getting info for unloaded ext…

sanmadjack:
This isn't a complete pull request, but an example of an idea I had to have a standardized way of providing information about an extension, even if it is not loaded. This would handle #525. Please review and let me know if this would be an acceptable way to handle this.

Also renames is_live to is_supported, since it has nothing to do with whether an extension is live.


shish:
Having extension metadata as code instead of trying to parse a comment with regex seems very sensible. Some thoughts:

- Code vs data - would something like JSON be more appropriate for metadata? The only reason I can see to have it as code is so that `is_supported()` can perform arbitrary checks (not just "what database are we using?" but "what third party extensions are installed?", "what OS is this?") - having metadata as JSON would require the framework to supply those functions. Not sure which approach is cleaner there.
- With opcode caches, I wonder if importing php files is actually faster than parsing json files...?
- If we're using code, should it be a separate class in `info.php`, or included in extension in `main.php`? If the latter, then we'd need to load every /main.php, and thus use some other mechanism to determine if an extension is active instead of `class_exists`...

So overall having a separate `info.php` for every extension seems reasonable :)

(Also for sake of simplicity and future maintenance, I'd migrate all the extensions to use that, instead of supporting both `info.php` AND regex'ed comment strings; though if you just want to get the first version landed which supports both, then I'm happy to do the work of migrating the rest and then dropping support for the old approach)

sanmadjack:
Well I googled around and it would that it if we used json files, opcode caches would not prevent them from being re-read every request, but keeping them in-php would allow the information to stay cached. If I somehow made the information constant, it may even further improve the caching efficiency. Since there are a good number of plugins at this point, I think preventing 30+ file reads every request would overall be beneficial. 

The whole class-isn't-loaded-for-disabled-extension functionality is something I think should be preserved, helps enforce the separation of the plugins. Keeping the info as a separate file gives us the best of both worlds.

As far as keeping things clean, I think if we make a generic info class to inherit from that has the actual is_supported functions, and the actual implementing classes basically just provide the raw data for the processes, it would stay pretty clean and organized. I get that json metadata is a very standard concept, but I don't know that it really gets us any benefit in this context. 

I'm going to finish setting up the working implementation of this, and I'll likely commit extension updates in bursts as I am able to. As you said, once they're all done we can just remove the old code at that point.

shish:
Yeah, that sounds like a good plan ^_^

sanmadjack:
Ci seems to be failing saying that the database socket can't be found. Did anything change recently?

shish:
Looks like something changed on the Travis end, since all the branches started failing a few days ago, even those which haven't been touched for months...

sanmadjack:
Checked in a fix for it. Don't know for sure what they changed, but I had to add mysql and pgsql as services. I also bumped the php version to 7.3, since that seems to be the target version now as it's what's available in debian stable.

sanmadjack:
This is failing due to some fragments from my code for pull request #670, which is itself pending review of the per-user settings feature that is checked in somewhere. I could go through and get the references removed, but if there's a chance that #670 will be accepted I'd rather not.

shish:
I've been procrastinating way too much over tiny details - I'mma accept as much as I can and then fix-forwards ^^

PR #683, opened 2019-09-14
Only show 'Tags' block when there's a tag to put in it

DanielOaks:
When you're using the `tag_categories` extension, you can get a case where the only tags that are shown are all in the category blocks and the main **Tags** block is empty, which looks weird. Seems like it's better to leave it out if that's the case, so that's what this change does \o/


shish:
Seems sensible :D

PR #684, opened 2019-09-16
Minor danbooru2 theme fixes

DanielOaks:
Just updates some stuff with the danbooru2 theme, basically:

Before, when you searched for a term, the result image links were a bit mangled. e.g. if I searched for the term 'wings', the result link would be `http://192.168.0.252:8123/index.php?q=/post/view/81search=wings`, dodgy because the image ID `81search=etc` doesn't exist (the stock theme returns `/post/view/81#search=wings` as expected). I've just made this theme do what the default theme does here and it works a lot better without any regressions or other breakages from what I can see.

Header didn't properly make the top parent link 'tab' blue, now it does so and makes the theme feel a bit nicer to use. What the header now looks like with this change applied:
Screen Shot 2019-09-16 at 11 12 36 am
Screen Shot 2019-09-16 at 11 12 42 am

### DIFF
--- a/themes/danbooru2/index.theme.php
+++ b/themes/danbooru2/index.theme.php
@@ -7,31 +7,15 @@ class CustomIndexTheme extends IndexTheme
      */
     public function display_page(Page $page, array $images)
     {
-        global $config;
-
-        if (count($this->search_terms) == 0) {
-            $query = null;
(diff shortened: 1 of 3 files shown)


shish:
Sorry for the delay, I wanted to merge the other big PRs first since they touched this file too ^^

PR #686, opened 2019-10-10
Cron upload

sanmadjack:
I implemented a few of the to-dos and buffed up the overall security and reliability of the process


shish:
All looks good ^__^

PR #687, opened 2019-10-10
Various changes

sanmadjack:
Another batch of changes and fixes for you to consider

### DIFF
--- a/core/polyfills.php
+++ b/core/polyfills.php
@@ -281,6 +281,7 @@ const MIME_TYPE_MAP = [
     'mpg' => 'video/mpeg',
     'mpeg' => 'video/mpeg',
     'mov' => 'video/quicktime',
+    'flv' => 'video/x-flv',
     'php' => 'text/x-php',
     'mp4' => 'video/mp4',
     'ogv' => 'video/ogg',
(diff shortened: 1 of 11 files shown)


core/polyfills.php
```diff
@@ -281,6 +281,7 @@ function mb_strtolower($str)
     'mpg' => 'video/mpeg',
     'mpeg' => 'video/mpeg',
     'mov' => 'video/quicktime',
+    'flv' => 'video/x-flv',
```
shish: flv is already defined as video/x-flv on line 261

ext/approval/info.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/*
+ * Name: Trash
+ * Author: Matthew Barbour <matthew@darkohlme.net>
+ * Description: Provides "Trash" or "Recycle Bin"-type functionality, storing delete images for later recovery
+ * Documentation:
```
shish: These seem to need updating~ (Do we even need metadata in comments, now that we have metadata in `ExtensionInfo`?)

shish:
Reviewed / tested / accepted most of these; a couple of minor nits and merge conflicts with what's left :3

ext/approval/info.php
sanmadjack: I would think not. We can probably just go ahead and delete them all.

core/polyfills.php
sanmadjack: Weird, it was missing in my copy. Must have been a merge mixup somewhere.

sanmadjack:
I re-based and addressed the noted issues, and tossed on a few more commits

PR #697, opened 2020-03-02
Auto-tag extension and small changes

sanmadjack:


shish:
All looks good - and with passing tests and everything :D

PR #701, opened 2020-03-18
Fixed Admin permissions

Mik-chan:
Found out that I'm unable to change image relationships due to some problem. I found out that it was because admins`Permissions::EDIT_IMAGE_RELATIONSHIPS` wasn't `true`. So I decided to set all admins permissions to `true` except Hellban.
I hope it would be useful.


shish:
Looks good, thanks for catching that :)

PR #705, opened 2020-03-23
Extend Wiki integration a bit

DanielOaks:
This colours wiki titles using tag categories, and also adds 'shortwiki' entries when viewing a single tag.

On some image board software, when you're only searching for a single tag it'll show you a short blurb about that tag. How this works in Shimmie with this PR is that you enable the _"Show shortwiki entry when searching for a single tag"_ checkbox in the board settings screen, then click on a tag. If no entry exists yet it'll just show you the title with a link to setup an entry. If an entry does exist it'll show you that plus the first line of that entry (just splits on newline and grabs the first line).

Here's how it shows up on the Danbooru2 and Default themes:
Screen Shot 2020-03-23 at 10 47 51 pm
Screen Shot 2020-03-23 at 10 47 44 pm

### DIFF
--- a/ext/index/theme.php
+++ b/ext/index/theme.php
@@ -37,6 +37,8 @@ and of course start organising your images :-)
      */
     public function display_page(Page $page, array $images)
     {
+        $this->display_shortwiki($page);
+
         $this->display_page_header($page, $images);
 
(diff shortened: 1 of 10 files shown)


shish:
LGTM ^_^

PR #714, opened 2020-03-28
Fix dirname(NULL) error in get_base_href

DanielOaks:
So this is an issue that only just started happening. A few days ago in git master, nope. Today in git master, this error is sprinkled through my docker logs on almost every page grab:

```
[Sat Mar 28 00:55:54 2020] PHP Fatal error:  Uncaught TypeError: dirname() expects parameter 1 to be string, null given in /app/core/polyfills.php:402
Stack trace:
#0 /app/core/polyfills.php(402): dirname(NULL)
#1 /app/ext/autocomplete/theme.php(7): get_base_href()
#2 /app/ext/autocomplete/main.php(67): AutoCompleteTheme->build_autocomplete(Object(Page))
#3 /app/core/send_event.php(113): AutoComplete->onPageRequest(Object(PageRequestEvent))
#4 /app/index.php(95): send_event(Object(PageRequestEvent))
#5 /app/tests/router.php(33): require_once('/app/index.php')
#6 {main}
  thrown in /app/core/polyfills.php on line 402


  [Sat Mar 28 00:56:06 2020] PHP Fatal error:  Uncaught TypeError: dirname() expects parameter 1 to be string, null given in /app/core/polyfills.php:402
Stack trace:
#0 /app/core/polyfills.php(402): dirname(NULL)
#1 /app/core/urls.php(34): get_base_href()
#2 /app/ext/et/main.php(32): make_link('system_info')
#3 /app/core/send_event.php(113): ET->onUserBlockBuilding(Object(UserBlockBuildingEvent))
#4 /app/ext/user/main.php(372): send_event(Object(UserBlockBuildingEvent))
#5 /app/ext/user/main.php(99): UserPage->show_user_info()
#6 /app/core/send_event.php(113): UserPage->onPageRequest(Object(PageRequestEvent))
#7 /app/index.php(95): send_event(Object(PageRequestEvent))
#8 /app/tests/router.php(33): require_once('/app/index.php')
#9 {main}
  thrown in /app/core/polyfills.php on line 402
```

This seems to fix it but I don't know enough about PHP to know if this is the _right_ way to fix it
¯\\\_(ツ)\_/¯

**edit:** for reference, the path to my install is http://localhost:8123/index.php?q=/post/list , 100% standard docker install on my macOS machine.

### DIFF
--- a/core/polyfills.php
+++ b/core/polyfills.php
@@ -391,7 +391,7 @@ function get_base_href(): string
         return BASE_HREF;
     }
     $possible_vars = ['SCRIPT_NAME', 'PHP_SELF', 'PATH_INFO', 'ORIG_PATH_INFO'];
-    $ok_var = null;
+    $ok_var = "";
     foreach ($possible_vars as $var) {
         if (isset($_SERVER[$var]) && substr($_SERVER[$var], -4) === '.php') {
             $ok_var = $_SERVER[$var];


shish:
I thiiiink this should be fixed in the latest docker image - I could see the problem happening with `docker run shish2k/shimmie2` (ie whichever build I pulled last, probably a day or so ago), but then running a local build didn't show it, and running `docker pull shish2k/shimmie2 && docker run shish2k/shimmie2` doesn't show it any more...

DanielOaks:
Ah I'll clear all my images and give both the shish2k image and from source a shot. I usually build from source so maybe my debian images are just acting weird~

DanielOaks:
Both latest source and the latest shish2k image throw the above error for me. I've spent a while tracking it down, and for me the commit 5ea7cc5b breaks it. If I `git checkout` efde5e1e and build+run the docker image there's no error, and if I `git checkout` 5ea7cc5b and build+run the docker image it throws the error. Given the change that commit makes to the router the dockerfile uses, makes sense that this'd start throwing errors I think.

shish:
Ack, you are right on both counts - the main thing I'm trying to figure out is why does removing PHP_SELF have any effect - I'm calling phpinfo() in a few places and it shows up as set-correctly-by-default o_O I am reproducing the issue consistently locally now at least /o/

shish:
Aha, found one thing at least: I thought it was being *randomly* triggered by /index.php - but it actually looks like it is *consistently* triggered by /favicon.ico (which is randomly-ish loaded by the browser whenever you visit /index.php)

shish:
wtf php

`/index.php?q=foo` -> `PHP_SELF = /index.php` <-- the documentation says that this is correct
`/foo` -> `PHP_SELF = /index.php/self` <-- no idea where this comes from
`/foo.bar` -> `PHP_SELF = /foo.bar` <-- shimmie complains because it can't locate index.php

shish:
OK so just in general the PHP built-in webserver treats "any path with a dot in it" and "any other path" via two very different processes. Even if the dot is in the middle, like /post/list/foobar/1 and /post/list/foo.bar/1 get processed differently -_-

shish:
So the end goal here is to have a function which looks at these possible situations, figures out what situation it is in, and then figures out that `base_href=/` and `query=post/list/foobar/1` (as opposed to `base_href=/post/` and `query=list/foobar/1`)

/post/list/foobar/1
Screenshot 2020-04-02 at 20 52 53
^ this case is easy, we just look for index.php, look at which folder it is in, and bam, that's our base_href

/post/list/foo.bar/1
Screenshot 2020-04-02 at 20 53 25
^ but then this

7045900503_5089ce25f1_z

shish:
Looks like everything is just terrible; 1d389f01 should at least put things back to how they were before :/

DanielOaks:
<3

PR #720, opened 2020-05-07
Added a new search criteria "fullrandom"

MetallicAchu:
The new criteria falls under `order[:|=]` and uses today's date as a seed for randomized list.
Thus the list is changing it's order every day, without any human interaction, giving a more dynamic feel to the imageboard.

### DIFF
--- a/ext/index/main.php
+++ b/ext/index/main.php
@@ -250,7 +250,15 @@ class Index extends Extension
             $seed = $matches[1];
             Image::$order_sql = "RAND($seed)";
             $event->add_querylet(new Querylet("1=1")); //small hack to avoid metatag being treated as normal tag
-        }
+        } elseif (preg_match("/^order[=|:]fullrandom$/i", $event->term, $matches)) {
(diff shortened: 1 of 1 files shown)


ext/index/main.php
```diff
@@ -250,7 +250,15 @@ public function onSearchTermParse(SearchTermParseEvent $event)
             $seed = $matches[1];
             Image::$order_sql = "RAND($seed)";
             $event->add_querylet(new Querylet("1=1")); //small hack to avoid metatag being treated as normal tag
-        }
+        } elseif (preg_match("/^order[=|:]fullrandom$/i", $event->term, $matches)) {
```
shish, line 253: "fullrandom" seems a weird name to me, like it implies that it's fully random? For something which is random based on the day, I'd call it "daily random" or something like that o.o

MetallicAchu: Yeah daily random sounds good :)

MetallicAchu: I suggest to change it to "dailyshuffle" because it makes more sense.
The order changes daily, so random doesn't describe it properly

PR #728, opened 2020-06-02
Various fixes and changes

sanmadjack:


sanmadjack:
Rebased on the two you cherry picked, and added some more things I've been working on

shish:
Now looking at the rest~

(I didn't pick the SQL fix yet because I thought that I'd already fixed that lower in the stack (so that everybody would get this behaviour by default), and I'd like to figure out what I'm missing from that code...)

core/basepage.php
```diff
@@ -257,8 +255,25 @@ public function display(): void
         } else {
             print "Error: Headers have already been sent to the client.";
         }
+    }
+
+    public function flush_output(): void
+    {
+
+    }
```
shish: This seems an unused stub? I guess it was added here first, then moved to polyfills.php, but not removed from here?

shish:
One of the original design decisions which has annoyed me forever was basing things on the file extension instead of the MIME type - thanks for moving us closer to the latter :D

sanmadjack:
> Now looking at the rest~
> 
> (I didn't pick the SQL fix yet because I thought that I'd already fixed that lower in the stack (so that everybody would get this behaviour by default), and I'd like to figure out what I'm missing from that code...)

I believe I made this fix quite awhile ago, I'm just getting it committed now. It's possible you did fix it and this is unnecessary.

core/basepage.php
sanmadjack: Yeah, I think that's what happened. I' removed it from my code, i'll commit it to this request.

shish:
Sounds good to me :)

I've merged a bunch - still to-do are the SQL fix (I want to figure out what was missing from my earlier attempt at a fix) and the two new extensions (I want to make sure I 100% understand them, so taking time to be extra-thorough)

sanmadjack:
It'ls possible the fix isn't needed anymore, I did make it quite awhile ago.

sanmadjack:
@shish  something's up with the postgres test, it's failing in the database update step. Do you have any idea what's causing it?

shish:
Merged a bunch of commits, and I've also fixed the postgres tests (github changed their test infra behind the scenes) - can you rebase so that we can see all tests passing on the still-to-be-merged parts? :D

sanmadjack:
Rebased, and 2 more commits at no extra cost.

core/imageboard/image.php
```diff
@@ -221,37 +221,22 @@ public static function count_tag(string $tag): int
     public static function count_images(array $tags=[]): int
     {
         global $cache, $database;
-        $tag_count = count($tags);
```
shish, line 224: These can reduce a ton of database load - I wonder if we can keep them, and skip when needed (eg, have some way for extensions to flag "I will alter the default search results") - or have them hidden behind the `SPEED_HAX` flag since that's defined as trading correctness for performance?

shish:
Merged a whole bunch more of these~

core/imageboard/image.php
sanmadjack: I agree, it's not optimal for performance, and since these extensions filter differently per-user, setting up new caching for it doesn't really make much sense. I like the speed hax idea, I can implement that.

sanmadjack:
Re-based

sanmadjack:
Tests are green once again.

shish:
Merged another whole bunch - did an interactive rebase to merge the adjustment commits in with their original commits; hopefully not too many conflicts now ^^

sanmadjack:
Rebased again. Is there an issue with the mime tests?

shish:
> Is there an issue with the mime tests?

Asking why I didn't merge `tests/test.{mov,asx,etc}`? I was just mildly concerned about the size (Test data for this one function is ~10x the size of the rest of the repository combined). Disk space is cheap, but it made my `git checkout` / `clone` / etc take a couple of seconds instead of a few milliseconds :P

Having an example file for each format that Shimmie can handle does seem like it'd be useful though 🤔

I wonder if there's a way for imagemagick / ffmpeg to generate example files in a scripted way? That'd be sliiiightly less reliable than having exact files committed as part of the repo; but IM and ffmpeg are fairly trustworthy, and that'd be more flexible...

(Also I did find one issue other than size - on OSX, .ico gets detected differently - fixed in 3297e716)

sanmadjack:
Fair enough. I'll rig up some media generators.

sanmadjack:
I added that SPEED_HAX criteria for the counts that we talked about, and I removed all the test files until I have the time to rig up generators.

sanmadjack:
I rebased and added a new commit so that the video media check function can also correct a matroska mime type to a webm mimetype when appropriate

sanmadjack:
I'm going to break this down into individual pull requests as much as I can

PR #730, opened 2020-06-13
Fixed imagemagick thumbnails for transparent images turning black

Ricardo50:
Thumbnails for transparent pictures used to be white in previous versions. Black makes it look ugly.


sanmadjack:
I've got a forthcoming commit that lets you pick the color from the settings screen.

shish:
Closing this - it's good, but the alternative fix which turns it into an option seems better - thanks for pointing out that it's an issue that needed fixing :D

PR #732, opened 2020-06-14
Tag EditCloud: Added category sorting and grouped tags.

LaureeGrd:
I'm hoping for this to serve as a starting point for a better implementation.

This change implements a simple **category-based alphabetical tag sorting** system that puts all tags containing ':' in front of general tabs. It also groups them together for easier styling into columns, grids, or even opening the door for drop-down categories in the edit menu. A much needed feature for me since I have hundreds of tags and I manage them all by my own.

Both features can be easily implemented on their own, although it's just a single commit because the styling feature only really benefits from category sorting (unless someone has a better idea for it).

EDIT: Haven't put my credits on it. I didn't feel like sticking my name on the side of someone that has made the entire extension on their own :P but feel free to add me if merged. I won't feel guilty that way.

### DIFF
--- a/ext/tag_editcloud/main.php
+++ b/ext/tag_editcloud/main.php
@@ -32,7 +32,7 @@ class TagEditCloud extends Extension
 
     public function onSetupBuilding(SetupBuildingEvent $event)
     {
-        $sort_by = ['Alphabetical'=>'a','Popularity'=>'p','Relevance'=>'r'];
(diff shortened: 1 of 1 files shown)


shish:
Sorry for the delay, life and pandemic have been a pain - the code looks good (the failing tests are due to github infra changing behind the scenes, I'm fixing that separately), so merging now, and happy to give credit :)

PR #737, opened 2020-07-17
Fix Notes system by delaying load of javascript

nateinu:
Notes system on images hasn't been working since upgrade from 2.6 to 2.8. Not sure why this works, but I added a delay to the loading of the ext/notes/lib javascript and commented out $('#main_image').off('mouseenter mouseleave');. Everything working as expected now.


shish:
That is interesting - Thanks for the pointer, I'd like to poke around and figure out _why_ that works before merging ^^;

PR #740, opened 2020-08-03
Added a list that sorts by total views

MetallicAchu:
Most of the code is stolen from numeric_score and ported in order to be usable here
I'm using the "image_views" table and cross-checked it with "images" to make sure that no deleted images are shown.
Showing images based on permissions, so every user sees what it's supposed to

TODO:

Invoke paginator - for now the list is single-paged only. Meaning that in theory thousands of pictures could be listed.

### DIFF
--- a/ext/image_view_counter/theme.php
+++ b/ext/image_view_counter/theme.php
@@ -1,0 +1,71 @@
+<?php declare(strict_types=1);
+
+class ImageViewCounterTheme extends Themelet
+{
+    public function view_popular($images)
+    {
+        global $page, $config;
+        $pop_images = "";
+        $i = 0;
+        foreach ($images as $image) {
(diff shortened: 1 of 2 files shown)


ext/image_view_counter/theme.php
```diff
@@ -0,0 +1,71 @@
+<?php declare(strict_types=1);
+
+class ImageViewCounterTheme extends Themelet
+{
+    public function view_popular($images)
+    {
+        global $page, $config;
+        $pop_images = "";
+        $i = 0;
```
MetallicAchu, line 9: This is probably redundant

shish:
Looks like this was never finished, but the idea is good - so I've manually merged and cleaned up what was there (and cleaned up a lot of what was there to start with, looks like there was quite a lot of redundant stuff ^^; )

PR #743, opened 2020-08-27
Fix child=any search not working

LaureeGrd:
Correct me if I'm wrong with this change, please.
But I found out child=none search was working, so I assumed there was an error in the check. This fixed it for me.

### DIFF
--- a/ext/relationships/main.php
+++ b/ext/relationships/main.php
@@ -83,7 +83,7 @@ class Relationships extends Extension
             }
         } elseif (preg_match("/^child[=|:](any|none)$/", $event->term, $matches)) {
             $not = ($matches[1] == "any" ? "=" : "!=");
-            $event->add_querylet(new Querylet("images.has_children $not :true", ["true"=>true]));
+            $event->add_querylet(new Querylet("images.has_children $not :true", ["true"=>"Y"]));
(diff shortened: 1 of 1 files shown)


shish:
This would work for databases where booleans are stored as Y/N - the database abstraction layer is _supposed_ to translate true/false into Y/N for those databases already though :S (and leave it alone for databases who natively support boolean types), so I'd wonder why is that not happening at the database layer...

LaureeGrd:
So, is there a possibility that this is an error on my side? Maybe I misconfigured something in the db...

shish:
I am pretty sure it's not your fault, just the database code not auto-converting `true` to `"Y"` when it's supposed to :)

Weird that unit tests are passing before and after this fix though - I would have expected them to fail on mysql (I presume?) before this PR, and fail on Postgres after it. I guess the tests probably don't cover this particular case >.<

shish:
This *should* be fixed at the database level now (and there are unit tests which show child=any working on all the databases) - feel free to reopen if it's still an issue with the latest code in the master branch :)

shish:
(Also note you'll need to run `composer update` to pull the latest libraries, as the fix was upstream)

PR #750, opened 2020-08-28
Extend autocomplete

sanmadjack:


sanmadjack:
@shish do you have some wariness/feedback on this?

shish:
Mostly I'm thinking about the half-done API things we've got so far (Danbooru, Ouroboros, Shimmie JSON, Shimmie Internal), wary of adding another new interface which follows new patterns, and only covers some parts of the database. I keep thinking "could there be a single API which Just Works and covers all the database structures with minimal effort for supporting new data types".

GraphQL looks like it could work really nicely as a universal query format, but I've not had time to investigate how hard it would be to make work with shimmie's existing data structures >.<

sanmadjack:
I get that. I was hoping the one I added could serve as the basis for a final solution, but obviously don't have the time to implement in any manner more than as needed. I was planning on extending what I implemented to support everything that might need it, but it's very much an as-we-go plan. 

Third-party APIs will always have their particular needs so there's not much to be done about them, but I think a single standard shimmie API is not a bad idea. If we do proper MVC stuff then parts of it could even act as a unified backend for the other APIs. I'll open a new issue to discuss this.

That being said, I can adjust this so that it just uses the tag data source that it was using previously.

shish:
Closing as things have changed so much - we now have autocomplete for nearly all tag inputs, based on `class="autocomplete_tags"` rather than any other kind of magic, and we also have a GraphQL API

Issue #759, opened 2020-10-13
Aspect Ratio Search Fails with SQL Error


jellykells:
**Server Software**

 - Shimmie version: 2.8.4+
 - Database: postgres
 - Web server: nginx

**What steps trigger this bug**
1. Try to search by aspect ratio as described in the included help page e.g. `ratio=16:9`

**What did you expect to happen?**
Filtered image results to appear normally.

**What actually happened?**
An error was returned:
```
Message: SQLSTATE[42725]: Ambiguous function: 7 ERROR: operator is not unique: unknown / unknown LINE 2: ... false OR owner_id = $1) AND (width / height > $2 / $3) AND ... ^ HINT: Could not choose a best candidate operator. You might need to add explicit type casts.

Query:

Version: 2.8.4+ (on 7.3.22-1+ubuntu18.04.1+deb.sury.org+1)

Stack Trace:

#0 /home/shimmie/shimmie2/core/database.php(262): Database->execute('SELECT COUNT(*)...', Array)
#1 /home/shimmie/shimmie2/core/imageboard/image.php(253): Database->get_one('SELECT COUNT(*)...', Array)
#2 /home/shimmie/shimmie2/core/imageboard/image.php(275): Image::count_images(Array)
#3 /home/shimmie/shimmie2/ext/index/main.php(76): Image::count_pages(Array)
#4 /home/shimmie/shimmie2/core/send_event.php(113): Index->onPageRequest(Object(PageRequestEvent))
#5 /home/shimmie/shimmie2/index.php(82): send_event(Object(PageRequestEvent))
#6 {main}
```

PR #761, opened 2020-10-14
[WIP] fix for failing ratio searches in postgres

shish:
step 1: add a unit test which demonstrates the failure


shish:
Context: #759 (this branch should fail under postgres but work in other places)

PR #807, opened 2020-11-15
[WIP] JSON-RPC API proof of concept

shish:
`POST` to `/api/json-rpc`, or command line `json-rpc <method> <params>`

Adding API support into existing extensions is pretty trivial, eg with the `autocomplete` extension already having a `complete(string $search, int $limit): array` function defined, we can add:

```
    public function onApiRequest(ApiRequestEvent $event) {
        if ($event->method == "autocomplete") {
            $event->result = $this->complete(
                $event->params->search ?? "",
                $event->params->limit ?? 0,
            );
        }
    }
```

A valid response:
```
$ php index.php json-rpc autocomplete '{"search": "a"}'
{
    "jsonrpc": "2.0",
    "result": {
        "angel": "1",
        "aaaa": "1"
    },
    "id": 1
}
```

An error:
```
$ php index.php json-rpc autocomplete '{"search": 42}'
{
    "jsonrpc": "2.0",
    "error": {
        "code": -1,
        "message": "Argument 1 passed to AutoComplete::complete() must be of the type string, int given, called in \/Users\/shish2k\/Projects\/shimmie2\/ext\/autocomplete\/main.php on line 36",
        "data": null
    },
    "id": 1
}
```

(If we wanted to populate the error's `code` or `data` fields, we could throw and catch `ApiException` objects?)

### DIFF
--- a/ext/autocomplete/main.php
+++ b/ext/autocomplete/main.php
@@ -28,6 +28,16 @@ class AutoComplete extends Extension
         $this->theme->build_autocomplete($page);
     }
 
+    public function onApiRequest(ApiRequestEvent $event)
+    {
+        if ($event->method == "autocomplete") {
+            $event->result = $this->complete(
+                $event->params->search ?? "",
+                $event->params->limit ?? 0,
+            );
+        }
+    }
+
     private function complete(string $search, int $limit): array
     {
         global $cache, $database;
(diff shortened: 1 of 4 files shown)


shish:
Now that I look at it, that autocomplete result seems off - should be a list of (tag, number) pairs rather than a map of {tag: number} - and should be int rather than string... but that seems to be coming from the original function, to be fixed separately

shish:
Closing because I like graphql more

PR #808, opened 2020-12-01
Show tag info on wiki pages

LaureeGrd:
Hi, I made this small change back in september so that wiki pages for tags display the aliases and auto-tags (if enabled).
This is my first addition and it's the base to a bigger one I was working on, so there may be mistakes.

I am trying to get back into coding so suggestions are appreciated. Thanks!


ext/wiki/main.php
```diff
@@ -332,6 +346,78 @@ public static function get_page(string $title, int $revision=-1): WikiPage
...
+                    
+                    $template = str_replace("{autotags}", implode(", ", $f_auto_tags), $template);
+                } else {
+                    $template = str_replace("{autotags}", format_text($config->get_string(WikiConfig::EMPTY_TAGINFO)), $template);
+                }
+            }
+        }
+        
+        //Insert page body AT LAST to avoid replacing its contents with the actions above.
+        $formatted = str_replace("{body}", format_text($page->body), $template ?? "{body}");
```
shish, line 417: This looks like it gets applied to all wiki pages, not just tag pages - so even "Site FAQ" etc will be put into the "tag page" template?

shish:
Sorry for not seeing this sooner, missed the notification >.< Looks good overall, but it looks like it affects all wiki pages, not just tag pages? o.o

ext/wiki/main.php
LaureeGrd: I coded this months ago but I'm pretty sure I did it with that in mind. In fact, that is the only line that is outside of the "if tag exists" block.
IIRC I did this because the function gets called as a replacement of [what was there](https://github.com/shish/shimmie2/pull/808/commits/a17e2eca15c7187c96885c7b335701006d1edcfa#diff-478b29418b9aaaf7ae189941a14fbc1451c388e1b4ae97d52e2a3f434a404992L80-R80).

If you notice, the last parameter in the line you cited is actually checking for the template, and if it is not assigned, it assumes a template "{body}" which means you will only get the body as without this change.
I'm not saying this is great design, but it works. Also, the idea of replacing the body after everything else is to avoid {thingslikethis} written in the wiki page itself to be replaced hehe

If you prefer a different method of deciding what to format and what to keep tho, I can try to make the changes.

LaureeGrd:
> Sorry for not seeing this sooner, missed the notification >.< Looks good overall, but it looks like it affects all wiki pages, not just tag pages? o.o

I just remembered I've done it like this to avoid doing the database query on the theme php file. And it also opens the possibility of formatting other types of wiki pages without too much hassle.
For example, I might add a list of tag categories when looking at wiki/tag_categories.

But don't worry, as I stated on my previous comment, the template for pages not corresponding to a tag is "{body}" which has the same result as not having this change implemented.

PR #810, opened 2020-12-21
Option to display post tags and related tags at the same time

LaureeGrd:
It seemed strange that this wasn't a feature. I love both options so I made it :)


shish:
That does look sensible, merging now, thanks :D

PR #815, opened 2020-12-28
Add missing entries in bbcode docs

LaureeGrd:
I had to look through the code to find the tags I wanted, so here they are.
Probably still missing a few, tho.


shish:
Thank you :D Sorry for the delay, I have no idea why github randomly doesn't notify me about pull requests 🙃

PR #816, opened 2021-01-16
Various fixes and changes

sanmadjack:


sanmadjack:
@shish what do i do about the formatting test?

shish:
> @shish what do i do about the formatting test?

`php-cs-fixer fix` run in the project folder should auto-format everything according to the PHP standards :3

shish:
That sounds sensible - the more automation and works-by-default the better IMO. Ideally I never want to think about formatting - I want it to _be there_, so that all the code looks consistent, and bugs jump out because the code "looks wrong" - but I want the computer to be the one who takes care of doing the work :D

PR #818, opened 2021-02-09
Added 2 missing block ids on post/view

LaureeGrd:
Small change. There's probably more to add and I will push those on this same PR if I find them.
I'm working on a custom theme with responsive design and I needed to be able to easily select those unidentified blocks.


LaureeGrd:
A quick search using grep shows that some other extensions are creating blocks without headers or IDs, but since I don't use those extensions and don't know what they do, I have decided to leave them untouched, for now.

shish:
Looks good, merging :)

Issue #827, opened 2021-05-01
Theming 'select' elements


LaureeGrd:
#### The problem
This is an example of what's mentioned in #826.
TL;DR: Extensions are using different methods to generate similar code (selectors in this case), but all of them need to be extended by themes that want to make just a simple HTML layout change like adding a wrapper div.

---
#### The possible workarounds
I had to extend two extension themes (and still have to extend many more across all of the site's selectors), copying the code from the original and changing it slightly to include my wrapper element. Another option in this particular case would have been to use JS to add the wrapper maybe, or work with what's there (no wrapper) and try to have consistent CSS (which I already tried as my first 2 approaches at theming, and it was horrible, especially once I added responsiveness to the equation).
If there's any other workaround I missed it, but those will probably have very strong cons to them as well.

---
#### My workaround
As mentioned before, I followed the "shimmie approach" and extended the themes to override the methods that generated those selectors. Here is the comparison:

https://github.com/shish/shimmie2/blob/c4ae68fb471382db6350b82236cf0cc52a2cdf81/ext/pools/theme.php#L383-L390

My extended code including the wrapper div and an icon:
```php
public function get_bulk_pool_selector(array $pools): string
{
    $output = "
    <div class='select-container'>
        <i class='fas fa-ellipsis-h'></i>    
        <select name='bulk_pool_select' required='required'>
        <option></option>
    ";
    foreach ($pools as $pool) {
        $output .= "<option value='" . $pool->id . "' >" . $pool->title . "</option>";
    }
    return $output . "</select></div>";
}
```

And the second one:
https://github.com/shish/shimmie2/blob/c4ae68fb471382db6350b82236cf0cc52a2cdf81/ext/rating/theme.php#L46-L62

My code including the same changes as the other one:
```php
public function get_selection_rater_html(array $selected_options, bool $multiple = false, array $available_options = null): string
{
    $output = "
    <div class='select-container'>
        <i class='fas fa-ellipsis-h'></i>
        <select name='rating".($multiple ? "[]' multiple='multiple'" : "' ")." >
    ";

    $options = Ratings::get_sorted_ratings();

    foreach ($options as $option) {
        if ($available_options!=null && !in_array($option->code, $available_options)) {
            continue;
        }

        $output .= "<option value='".$option->code."' ".
            (in_array($option->code, $selected_options) ? "selected='selected'": "")
            .">".$option->name."</option>";
    }

    return $output."</select></div>";
}
```

---
#### The proposed solution
[...]

PR #828, opened 2021-05-03
Standard HTML select code

luanalatte:
Same proof of concept from #827 but without my custom wrapper. `build_selector` has been moved to `basethemelet.php`. I didn't even think of it before. Putting this function here means that in addition to overriding on the entire site, themes can override them on a per-extension basis, giving even more control to the designer.
Let's just hope we don't override the new functions on extensions themselves or we will end up in the same road :smile:

- The doc-string might be a bit too much, I just wanted to make it clear for this PR.
- There's an unrelated change that makes pools sort posts by ID when order is the same. It was a personal uncommitted change and it slipped on this one.

If this looks good, I will implement `build_selector` on more extensions.

### DIFF
--- a/core/basethemelet.php
+++ b/core/basethemelet.php
@@ -196,8 +196,9 @@ class BaseThemelet
      * Generates a <select> HTML template and sets up the given options.
      *
      * @param string $name The name attribute of <select>.
-     * @param array $options A pair of parameters for <option> tags. First one is value, second one is text. Example: ('optionA', 'Choose Option A').
-     * @param string $attributes Flags for <select>. Example: 'multiple required' becomes <select multiple required>
+     * @param array $options An array of pairs of parameters for <option> tags. First one is value, second one is text. Example: ('optionA', 'Choose Option A').
+     * @param bool $required Wether the <select> element is required.
+     * @param bool $multiple Wether the <select> element is multiple-choice.
      * @param bool $empty_option Whether the first option should be an empty one.
      * @param array $selected_options The values of options that should be pre-selected.
      */
--- a/core/basethemelet.php
+++ b/core/basethemelet.php
@@ -202,7 +202,7 @@ class BaseThemelet
      * @param bool $empty_option Whether the first option should be an empty one.
      * @param array $selected_options The values of options that should be pre-selected.
      */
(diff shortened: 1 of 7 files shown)


shish:
This does look good :)

To deal with the formatting test:
```
./vendor/bin/php-cs-fixer fix
```

LaureeGrd:
> This does look good :)
> 
> To deal with the formatting test:
> 
> ```
> ./vendor/bin/php-cs-fixer fix
> ```

Thanks. I'll fix everything when I feel like it's ready for full review and merging.

luanalatte:
@shish I decided to work on this old improvement I never got to finish. Sorry for the force-push, I think it's safe to assume no one else had this cloned :P

Quick Recap:
- Extensions should use standard code for generating HTML where possible, and since `<select>` tags are widely used on shimmie, it's a good place to start. `build_selector` in `core/basethemelet.php` takes care of that.
- The changes I did in the `pools` and `ratings` extensions are just examples, other extensions are still pending to implement this function.
- I'm using MicroHTML this time, but since it is my first time working with it, there may be better ways to do it.

Note: I used named parameters from PHP 8, if that's not your style we can work around that.

shish:
That does look nicer :D


> Note: I used named parameters from PHP 8, if that's not your style we can work around that.

Oooh, I hadn't noticed that we can start using those now that PHP7 is dead, I'd be happy to have more of that :)

shish:
For the functions themselves - I think it'd be nice to have all the standard HTML-generation functions returning `HTMLElement` rather than `string`, as manipulating HTML as strings is such a security risk (I'll also look at updating the existing themelet functions, as they are currently setting a bad example...)

luanalatte:
Ok, I fixed the small conflict and (sort of) made it work with `HTMLElement`s instead of strings. This PR shows how much of a noob I still am working with git branches, lol. I guess next time I'll prefer having an extra "merge" commit from master rather than ruining the PR's history like this.

There's something I'd like to review before continuing implementation, though. I'll open a discussion on it if I remember how it's done...

shish:
Not sure what was up with github actions, but I've retried now and they all passed /o/

Merging now so that we can work in parallel and avoid merge conflicts :3

PR #829, opened 2021-05-03
Attempt at externalizing layout from extensions

luanalatte:
Decided to go for a different branch for this, so we can easily discard it or merge onto the other one, since I'm still trying to find the workflow we should go for with this rework.

If we want to let themes completely replace layouts, like replacing this table with a flexbox or grid, we should externalize the TR, TH, TD wrappers found in extensions. This is just one example of them.

I decided to go for a custom class for parts of this `ImageInfo` block but this is not the only place where an object with `header`, `body` and `order` is needed, so I can see Parts becoming its own thing like Blocks but with just the bare minimum information. They would be useful for events that build blocks like this one. Maybe we end up using an array and not a dedicated class for this, idk.

The part building process is not definitive, it's just a quick prototype.

I wanted to use MicroHTML on `ext/view/theme.php`'s `build_info` but the contents of the table still needed to allow for HTML to be injected. This may be improved as well.

There's a few TODOs here, mainly notes for myself. Also, sorry @shish, I know this is a lot of work and a lot to think about and review. I'm trying to maximize my contributions since this is my last free week before starting uni.

### DIFF
--- a/ext/setup/theme.php
+++ b/ext/setup/theme.php
@@ -89,7 +89,7 @@ class SetupTheme extends Themelet
     protected function sb_to_html(SetupBlock $block): string
     {
         $h = $block->header;
-        $b = $block->body;
+        $b = $this->build_setup_chunk($block->body);
         $i = preg_replace('/[^a-zA-Z0-9]/', '_', $h) . "-setup";
         $html = "
 			<section class='setupblock'>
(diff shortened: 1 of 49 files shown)


ext/setup/theme.php
```diff
@@ -99,4 +99,36 @@ protected function sb_to_html(SetupBlock $block): string
...
+        $output .= $html;
+        if ($table_row) {
+            $output .= "</td>";
+        }
+        if ($table_row) {
+            $output .= "</tr>";
+        }
+
+        return $output;
+    }
```
LaureeGrd, line 133: main.php still uses a lot of small non-overridable functions to generate tables (I think we should avoid that) but at least this one can be overridden now.

ext/view/theme.php
```diff
@@ -85,7 +85,17 @@ protected function build_info(Image $image, $editor_parts): string
 					<table style='width: 500px; max-width: 100%;' class='image_info form'>
 		";
         foreach ($editor_parts as $part) {
-            $html .= $part;
+            // This could be using something like MicroHTML? but the contents need to allow HTML anyway.
```
shish, line 88: You can add unsafe generic HTML into the middle of an otherwise-safe structure with rawHTML like

```
use function MicroHTML\rawHTML;
[...]
TH(rawHTML($part->header))
```

This also makes it easy to search for `rawHTML`, so if we see some invalid HTML being generated somewhere, we can focus on those parts :)

ext/tag_edit/theme.php
```diff
@@ -69,20 +62,12 @@ public function get_user_editor_html(Image $image): string
...
 					<span class='view'><a class='username' href='".make_link("user/$h_owner")."'>$h_owner</a>$h_ip, $h_date</span>
 					<input class='edit' type='text' name='tag_edit__owner' value='$h_owner'>
 		" : "
 					<a class='username' href='".make_link("user/$h_owner")."'>$h_owner</a>$h_ip, $h_date
-		")."
-				</td>
-				<td width='80px' rowspan='4'>$h_av</td>
-			</tr>
-		";
+		"); //TODO: Implement removed <td width='80px' rowspan='4'>$h_av</td>
```
shish, line 70: This is the only thing stopping me from merging at the moment, as avatars are a pretty important part of user-identity ^^

I'd be tempted to implement it by adding it to `ext/view/theme.php` - it's not ideal, but it would be a small step back while everything else here is several steps forwards

ext/view/theme.php
LaureeGrd: That is nice. Will add this to my TODO list :)

ext/tag_edit/theme.php
LaureeGrd: oh, don't worry. I'll implement it accordingly. This PR is not meant to be merged yet as it is a bit less polished, because I'm working on multiple things at the same time. I wanted to make sure I wasn't making a big mistake and have to discard too much work. I'm working on setup's layout at the moment, which was a big pain when I tried to theme it.

ext/setup/theme.php
shish: FWIW the current approach of adding lots of TD/TH/TR elements seems flexible but messy - ideally this'd be another place where the extension just returns a list of `{setting_name, human_name, value, display_style}` and the theme takes care of HTML, but I'm not sure how to do that without massively reducing flexibility...

(This isn't something that needs fixing right away, just something that I have in the back of my mind)

LaureeGrd: It seems flexible right until you don't want to use tables :smile: 
Also, I'm noticing extensions create their tables in very different ways, which explains why I had so much trouble getting everything to look consistent on my theme. There's even setupblocks that don't use tables at all and just populate the items in the div.

I know it's going to look a lot less flexible when I commit this. But giving extensions that much flexibility leads to inconsistencies...
I'm also trying to send everything through this `format_option` function, which I now renamed to `format_item` and takes care of labels and text as well, and I created other functions in theme.php to wrap them in the table elements (the ones one would override to get rid of tables entirely)

We can discard it if you think it's too much anyway. I'm adapting all extensions to the revised format now, I'll probably commit after that.

ext/user_config/main.php
```diff
@@ -141,7 +141,7 @@ public function onPageRequest(PageRequestEvent $event)
                         return;
                     }
 
-                    $uobe = new UserOptionsBuildingEvent($display_user, new SetupPanel($user_config));
+                    $uobe = new UserOptionsBuildingEvent($display_user, new SetupPanel($user_config, new SetupTheme));
```
LaureeGrd, line 144: This is the only thing that's not working with the `SetupBlock` changes. How should I access `SetupTheme`? Maybe `UserConfigTheme` should extend `SetupTheme` now? I feel like it should.

ext/setup/theme.php
```diff
@@ -99,4 +99,56 @@ protected function sb_to_html(SetupBlock $block): string
...
+            $html = $this->build_setup_cell($html, $label_is_header, $full_width);
+        } else {
+            $html = "";
+        }
+
+        if ($label_is_header) {
+            return $label . $this->build_setup_row($html);
+        }
+        
+        return $this->build_setup_row($label . $html);
```
LaureeGrd, line 152: Found a bug here. I'll work on it tomorrow.
It's pretty harmless tho, creates an empty table row when formatting a full width label.

Issue #839, opened 2021-06-15
Sort tag edit cloud -- by category --> by sum of tags of category 


MetallicAchu:
When you enable tag categories, and choose to sort the tag edit cloud by categories, it sorts the categories by alphabetical order.
I want it to sort by the amount of tags in each category.
For example, if I have 2 categories:
Animals:cats - 10 tags
Animals:dogs - 5 tags
Object:Door - 20 tags
Object:Car - 30 tags

I want Object category to appear first because it has 50 tags.

I tried doing it myself but got into too much trouble formatting the SQL query :)

Thanks!

MetallicAchu:
Figured it out eventually
```
$tag_data = $database->get_all(
                        "
                                        SELECT tag, FLOOR(LN(LN(count - :tag_min1 + 1)+1)*150)/200 AS scaled, count
                                        FROM tags
                                        WHERE count >= :tag_min2
                                        ORDER BY SUM(count) OVER (PARTITION BY SUBSTRING_INDEX(tag, ':', 1)) DESC, CASE 
                                            WHEN tag LIKE '%:%' THEN 1
                                            ELSE 2
                                        END, tag
                                        LIMIT :limit",
                        ["tag_min1" => $tags_min, "tag_min2" => $tags_min, "limit" => $max_count]
                    );
```

Will open a pull request on this one

PR #840, opened 2021-06-17
Edit sort by Categories

MetallicAchu:
Sort by the number of tags (sum) for each category descending
This will allow the most used categories to appear first and not by alphabetical order


MetallicAchu:
#839 request fixed here

Issue #836, opened 2021-05-31
'Fit Height' and 'Fit Both' locks file dimensions


PVNFU-28:
**Server Software**
 - Shimmie version: 2.9.1+
 - PHP: 7.4.3
 - Database: MySQL 8.0.25
 - Web server: Apache 2.4.41

**Client Software (please complete the following information)**
 - Device: Windows laptop
 - Browser: Chrome

**What steps trigger this bug**
1. Go to https://carbu.tk/post/view/55
2. Click on 'Fit Height' or 'Fit Both'

**What did you expect to happen?**
I expected for the image to fit the height and then be able to change the dimensions by click another option like 'Fit Width' or 'Full Size'

**What actually happened?**
When you click any other option after clicking 'Fit Height' or 'Fit Both', the image will stay the same, to able to change the images size you have to reload the page.
I know this is not my installations fault because it also happens on other boorus running the same version, rule34.paheal.net for example. This does not happen on Shimmie Testbed I suspect because hes still on Shimmie version 2.8.4

MetallicAchu:
Recreated the same problem using multiple browsers.
I think this is a problem with how the cookies are saved

shish:
This should be fixed now? Feel free to re-open if not :)

PR #844, opened 2021-09-06
Add missing quote in html returned by handle_pixel theme

### LINKED ISSUE #825: "Fit Height" breaks image preview
I noticed this while theming, and tried it with the Default theme too.
On the three browsers I tried on desktop, changing this option to "Fit Height" leads to it not behaving correctly anymore (you can't change back to the others) until you reload the page. It may be happening with any of the other 3 options as well but I noticed it does this consistently with "Fit Height".
On android it worked a little better but I still see some weird behavior. I'm not able to point out what exactly.

I would simply disable this option from my theme but the function and script that add it cannot be overriden by themes, right?
https://github.com/shish/shimmie2/blob/c4ae68fb471382db6350b82236cf0cc52a2cdf81/ext/handle_pixel/main.php#L58-L72

This is not the only problem I noticed while theming but it is the only one (yet) that I would consider a bug. I will most probably be posting another issue with a more general list of things that we should address for theming.
But for now I'm curious to see if we can fix or circumvent this.

---
**Server Software**
 - Shimmie version: 2.9.1+
 - Database: mysql 10.5.9-MariaDB
 - Web server: lighttpd/1.4.59

 - Device: android, linux desktop
 - Browser: firefox, palemoon, qutebrowser, android's chrome, android's firefox nightly

1. Change the Fit option on the "Post Controls" section of an image to "Fit Height"
2. Change back to others
3. Now the fit option is frozen and you can't return to "Full Size" for example.

jellykells:
bug introduced in e82b9ea811

should fix #825, #836

### DIFF
--- a/ext/handle_pixel/theme.php
+++ b/ext/handle_pixel/theme.php
@@ -30,7 +30,7 @@ class PixelFileHandlerTheme extends Themelet
         }
 
         $html = "<img alt='main image' class='shm-main-image' id='main_image' src='$u_ilink' ".
-            "data-width='{$image->width}' data-height='{$image->height} data-mime='{$image->get_mime()}'>";
+            "data-width='{$image->width}' data-height='{$image->height}' data-mime='{$image->get_mime()}'>";
         $page->add_block(new Block("Image", $html, "main", 10));
     }
 }


shish:
Looks good, thank you <3

jellykells:
that was fast :D
seems like Github only autoclosed one of the linked issues though

PR #850, opened 2021-11-11
Added missing colon to version title on extension page

HeCorr:
Before:
version-before

After:
version-after


shish:
Nice catch, thanks :)

PR #851, opened 2021-11-21
Add option to return 404 code for nonexistent wiki pages

jellykells:
I noticed search engines indexing wiki pages on my booru that didn't exist since they return a 200 code. this option sends a 404 instead if the wiki page doesn't exist when it's enabled.

if adding the function to the WikiPage class isn't the preferred way to do this let me know how you'd prefer me to implement it

### DIFF
--- a/ext/wiki/main.php
+++ b/ext/wiki/main.php
@@ -50,11 +50,13 @@ class WikiPage
     public string $title;
     public int $revision;
     public bool $locked;
+    public bool $exists;
     public string $body;
 
     public function __construct(array $row=null)
     {
         //assert(!empty($row));
+        global $database;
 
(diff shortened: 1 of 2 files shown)


shish:
Sorry for the delay, notifications are failing for me >_<

I wonder if there's a reason to have this as an option? It seems like sensible behaviour all the time to me, I don't recall "missing pages should return HTTP 200" being a conscious design decision, I think it was an oversight ^^;;

jellykells:
I made it an option since it changes the default behavior, but if you think it's reasonable then I'll go ahead and remove the configuration option.

shish:
Looks great, thank you <3

PR #860, opened 2021-12-19
Fixing home page pool link and fixing Notes not displaying due to css

yls4:


yls4:
Not sure where the css for current notes extension comes from, the imgNotes repository had a different css

PR #863, opened 2021-12-24
Allow image owner to view their unapproved images

jellykells:
this allows a user to view the images they've uploaded, even if they haven't been approved. helpful to allow use of quality-of-life features like the tag edit cloud or correction of mistakes before approval

the owner may also comment on the unapproved image but these comments are hidden from the comment list until the image is approved


jellykells:
ah I broke a test :|

Issue #800, opened 2020-10-28
Make use of X-Forwarded-For


shish:
Is this actually a good idea? It's missing in proxies that care about privacy, and can be faked...

We could at least log it somewhere maybe?

dali99:
X-Forwarded-For is required when using a reverse-proxy in front an application. If you don't, all logged ips are just the reverse proxy

It should be turned on with a setting though, or have an ip whitelist for trusted proxies

PVNFU-28:
running shimmie2 with a reverse proxy and currently facing this issue, a setting would be great

shish:
I'd personally been using the X-Real-IP boolean setting from #869 for a while, but added c5395df to make use of either X-Real-IP or X-Forwarded-For, with a list of trusted proxies rather than a boolean

PR #869, opened 2022-01-20
X-Real-IP support and Varnish PURGE config options

thoughever:
Related to #800 and #862 

I added 2 features to make it easier to run Shimmie2 behind a reverse proxy.

1. Setting the global `REVERSE_PROXY_X_HEADERS` to true will get IPs from the `X-Real-IP` header and the protocol from `X-Forwarded-Proto`. I didn't use `X-Forwarded-For` because some proxy servers set it to a list of IPs which is somewhat ambiguous.
2. Config options to set the host that should receive Varnish PURGE requests

### DIFF
--- a/core/captcha.php
+++ b/core/captcha.php
@@ -11,7 +11,7 @@ function captcha_get_html(): string
 {
     global $config, $user;
 
-    if (DEBUG && ip_in_range($_SERVER['REMOTE_ADDR'], "127.0.0.0/8")) {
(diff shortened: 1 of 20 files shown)


shish:
Hi, sorry this took so long to get to - turns out that having COVID kind of sucks and will mess with a person's schedule ^^;; Code looks great though, and fixes an issue I was having myself /o/

shish:
FYI c5395df changes the REVERSE_PROXY_X_HEADERS boolean to a TRUSTED_PROXIES list of IP addresses

Issue #849, opened 2021-11-09
Can't run in Docker (Windows) from source


HeCorr:
**Server Software**
 - Shimmie version: `v2.9.1`
 - Database: none yet
 - Web server: the default (?)

**Client Software**
 - Device: Windows 10 Pro [1909]
 - Browser: not applicable

**What steps trigger this bug**
1. Clone the repo
2. Build the Docker image [as per instructions](https://github.com/shish/shimmie2/wiki/Docker#build-custom-image)
3. Try to run the container. I used the command `docker run --rm -ti -p 8000:8000 --name shimmie shimmie2`

**What did you expect to happen?**
I expected it to start normally.

**What actually happened?**
This error showed up:
```
' is not a valid group name
useradd: group '1000' does not exist
chown: invalid user: 'shimmie:shimmie'
su-exec: getpwnam(shimmie): Success
```

**Extra info**
First I tried building from the `master` branch, and when this error occurred, I checked out the latest release tag with `git checkout tags/v2.9.1 -b latest` and tried again, with no success (and yes I did use `--no-cache`).
Initially I tried running the image on a Linux server but the build process was taking so long that I moved it onto my Windows machine.
By the way, I **had** successfully run the official image from Dockerhub before, but I don't remember which version it was.

shish:
Ah, looks like upstream debian released a new version, and since shimmie builds on top of "debian:testing" (Which up until recently meant "the beta for v11" and now means "the beta for v12"), that also got bumped to a major new release. I expect the fix would be to update the Dockerfile to use `debian:stable` (ie, v11). I'm testing that now, but I'm on mobile internet so it'll take a while to download the new docker base images...

HeCorr:
Ohh, that would make sense. I'm trying it right now as well  ;)

HeCorr:
Aaaand nope, same error unfortunately.  :(

![image](https://user-images.githubusercontent.com/75134774/141002282-ce6c4985-216e-48ff-819f-e52d39b25374.png)

shish:
Hmm, weirdly it's working for me with either version...

Can you try adding at the top of tests/docker-init.sh
```
#!/bin/sh
echo UID=$UID, GID=$GID
```
And then rebuild and rerun? It looks like one of those is somehow set to something unusual (they should both be "1000" by default)

HeCorr:
Sure thing! gimme a sec.

HeCorr:
Okay, this is the result:

```
[...]

PR #870, opened 2022-02-10
Fix apt errors when running in Docker

HeCorr:
I was facing some errors when running Shimmie2 in a Docker container:

image

Which was fixed by replacing the `debian:testing-slim` source image with `debian:stable` in the Dockerfile, as suggested by @shish [here](https://github.com/shish/shimmie2/issues/849#issuecomment-964527949) (for a different issue lol).


shish:
Thank you :)

PR #874, opened 2022-04-29
Allow ordering by favourites in searches

ticky:
This allows sorting by the number of favorites a post has in the same way as other ordering metatags. If there's a better way to handle conditionally showing the metatag in autocomplete, let me know!

closes friends-of-the-core/shimmie2#1


shish:
LGTM, thank you ^__^

PR #875, opened 2022-04-29
Allow pre-setting `$image->posted` values

ticky:
This updates the `Image` class to permit setting `posted` before the item is committed to the database.

This allows for backdating (or, of course, forward-dating) uploads in custom extensions. An example of taking advantage of this is [in our fork](https://github.com/friends-of-the-core/shimmie2/blob/a1eb45ede98f1de8cec50c2618b32d70f1d98262/ext/filename_metadata/main.php) where we are ultimately taking the post time information (and user information) from a Discord channel.

I believe there may be a bug here preventing this from working with MySQL backends; I've only fully tested this against SQLite (which doesn't care) and Postgres (which perfectly accepts this date format). If you've got any hints for this, let me know!

Additionally, I added the ability to set the uploader in the same manner, but ended up not taking advantage of it in favour of tags. Feel free to cherry-pick if you're interested! https://github.com/friends-of-the-core/shimmie2/commit/5d2b835c262c55ac217dd422335cb46a65056229

closes friends-of-the-core#3

### DIFF
--- a/core/imageboard/image.php
+++ b/core/imageboard/image.php
@@ -28,7 +28,7 @@ class Image
     public ?array $tag_array;
     public int $owner_id;
     public string $owner_ip;
-    public string $posted;
+    public ?string $posted = null;
     public ?string $source;
     public bool $locked = false;
     public ?bool $lossless = null;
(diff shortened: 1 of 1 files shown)


shish:
Unit tests seem happy on all databases, so I'mma merge this as-is and take responsibility for adding more tests if mysql complains :)

PR #883, opened 2022-07-06
fix no handler for 'bulk_download'

JamesTheSergal:
Shimmie out of the box has bulk downloads enabled for the 'user' and 'base', but perform bulk actions isn't enabled, so it goes to a handler not found page. This fixes that.


JamesTheSergal:
This is the most simple thing I could have probably contributed to this project. But, I hope to contribute much more in the future, as I have been using this project on my own sites for years, I'd really love to help out.

PR #885, opened 2022-07-11
Autocomplete on upload page

JamesTheSergal:
This little change is thanks to https://github.com/ticky
It adds the autocomplete onto the upload page specific tag input form. I find it to be extremely useful.


tegaki-tegaki:
@JamesTheSergal This looks really good! (and so simple too, oneliner!). I would just do a linter fix so it looks "flawless", that could prevent a maintainer from looking into it. I can't see the logs for the format test, so 🤷 don't know why the test failed.

shish:
This should be working in the master branch as of a while ago, though it was done differently (making `class=autocomplete_tags` consistently enable autocomplete, rather than having a whole bunch of conditions to fulfil)

PR #888, opened 2022-09-27
Fixing bulk add bug, tag is set to "\" if no other tags are present

yls4:
Images will now be uploaded with no tags, which will cause them to be generated with "tagme" instead.
This function is also used by archiver, hopefully not a problem.


shish:
Seems legit, thanks for the fix :)

PR #892, opened 2022-11-13
Update memcache DSN condition matching

Cirvaazny:
Add an additional OR conditional for memcache usage. If you use the DSN as listed at https://github.com/shish/shimmie2/wiki/Performance#software-cache the memcache section will never run because it is looking only for DSNs with 'memcached' causing an exception to be thrown on the first access as the cache is never setup.


shish:
Looks good to me :)

(For some historic context - originally PHP had two different popular libraries for connecting to memcache - one named "memcache" and one named "memcached", and shimmie let you select which library to use. But now that there's only one supported library, it makes sense to have both names aliased to the same code.)

PR #895, opened 2023-01-10
Fix database not detected on installation

nodate-yaozua:
Hello! I'm new to Shimmie and today I tried running the latest version. Looks like some string conversions are missing from the install.php. Thank you!


shish:
I wonder how none of the automated tests caught that... Code looks good though, thank you! :D

PR #900, opened 2023-02-04
[tombstones] keep track of what gets deleted

shish:


AstroAntics:
Did you get the idea from Newgrounds? I remember they had a tombstone there too. I don't remember if it had a reason or not to go along with it.  Do you think there's any benefit to showing the user why the image was deleted?

I personally believe that there would be (i.e. "banned hash", "poor quality", "TOS", etc), but I can see the flip-side of the argument as well. Some other sites already do this anyway, and display why an image was deleted if you go over it.

shish:
I got the idea from end-user requests, no idea where they got it from - there are pros and cons, but if it's an option that site owners can choose to enable or disable for their own sites then it seems sensible to add it :3

PR #901, opened 2023-02-04
Start of ActivityPub support

shish:
Not even sure how this would work, just playing with the API


tegaki-tegaki:
Looking forward to this one btw! (can try it out when it's test-ready!), hosting some other fediverse stuff as well as shimmie, would be cool to "glue" them all together!

It's probably a very deep rabbit hole...

shish:
I'm currently not personally using any fediverse services, so I'm not even sure what the end result should look like - input of ideas and / or code from other people would be welcome ^^;;

PR #905, opened 2023-03-17
Fixes

NottyNoz:


ext/private_image/main.php
```diff
@@ -114,7 +114,7 @@ public function onDisplayingImage(DisplayingImageEvent $event)
     {
         global $user, $page;
 
-        if ($event->image->private===true && $event->image->owner_id!=$user->id && !$user->can(Permissions::SET_OTHERS_PRIVATE_IMAGES)) {
```
shish, line 117: Can you give an example of what problem this change is solving? I have a feeling that it probably solves one problem and creates a different one, so I'd like to understand it fully to make sure the original problem is fixed without side effects ^^

(for example, if `$event->image->private` is set to the string `"false"`, then that will be considered equal to the boolean `true`...)

NottyNoz: `$event->image->private` returns 1 or 0 which isn't identical (===) to true. This means that redirection never occurs and anybody with the url to the post can see it, even users without an account. I'm using PHP 8.1 and haven't tested this comparison in other versions. This comparison issue is such an obvious bug that it makes me think there was a change to PHP.

The trash extension also uses the same data type `Image::$bool_props[] = "trash"` and uses similar comparisons `$image->trash===true` however I haven't tested it.

shish: Seems legit - I'd like to make bools act consistently like bools instead of sometimes being ints, but for now this change seems at least a step in a more sensible direction :)

PR #908, opened 2023-03-22
Theme and UI Fixes

NottyNoz:
- Show important links in User Links on default theme
- Cleaned up and improved some extension documentation
- Fixed a bug with the Tag-it plugin where the wrong input box would get targeted on a page with multiple Tag-it input boxes
- Fixed Lite theme UI
- Fixed Futaba theme UI
- Added a file size tracker to the upload page

There should at least be a mention somewhere regarding post_max_size, it is as important as upload_max_filesize

### DIFF
--- a/ext/approval/main.php
+++ b/ext/approval/main.php
@@ -121,6 +121,13 @@ class Approval extends Extension
         }
     }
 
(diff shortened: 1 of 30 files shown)


shish:
Most of these commits got merged a while ago, but it is all merged now /o/

PR #909, opened 2023-03-26
Bulk Actions for setting a chain of parent child relationships in the…

yls4:
… order of images selected.

Does not support setting multiple children to one parent in bulk.


ext/bulk_parent_child/info.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
```
shish, line 2: These files need a `namespace Shimmie2;`

ext/bulk_parent_child/main.php
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class BulkParentChild extends Extension
+{
+    private const PARENT_CHILD_ACTION_NAME = "bulk_parent_child";
+
+    public function onInitExt(InitExtEvent $event)
+    {
+        global $config;
+    }
```
shish, line 19: This can be deleted

ext/bulk_parent_child/main.php
```diff
@@ -0,0 +1,51 @@
...
+            $event->add_action(BulkParentChild::PARENT_CHILD_ACTION_NAME, "Set Parent Child");
+        }
+    }
+
+    public function onSetupBuilding(SetupBuildingEvent $event)
+    {
+        $sb = $event->panel->create_new_block("Bulk Parent Child");
+        $sb->start_table();
+        $sb->end_table();
+    }
```
shish, line 35: This should also be deleted rather than adding an empty config block :)

shish:
Looks good, thank you :D

PR #917, opened 2023-05-24
add extension: hide_tags

tegaki-tegaki:
# why this extension

I made this feature for my booru so I could arbitrarily "soft hide" posts, while still making them visible if you have the post_id (eg if you link to an image hosted on the booru elsewhere). I wanted this because I wanted to essentially treat my booru _mostly_ like a _themed_ booru, but I also wanted the ability to treat it a bit like a generic image host, but if I post whatever on it, then people who come to the booru for the _theme_ will see all the _non-themed_ posts! This will not do! I know the "private" addon does the same thing but that opens the booru up to be more of a "private image host", and that's not quite what I want either, I just want to be able to casually hide images without too much fuss. It's not "well hidden", and it's not fully visible, just something in between :)

It can probably be used for other things too? 🤔 

I don't suggest merging it in _as is_, because It's not "clean enough" imo. I kinda "hacked it together" without too much thought. I'll try annotate the code with some PR comments of my own.

# features

1. if you tag a post with the now _special_ tag `hidden` the post will be hidden! (the way it does this is by injecting the string `-hidden` to all searches)
2. if you search for posts with the now _special_ tag `show_hidden`, it will let you search normally across all `hidden` tagged posts as well.

### DIFF
--- a/core/imageboard/image.php
+++ b/core/imageboard/image.php
@@ -9,6 +9,19 @@ use GQLA\Field;
 use GQLA\Query;
 
 /**
+ * placed here until it finds a better home
+ *
+ * something seems to be special about core extensions? (at least
+ * in regards to loading the config from a non-core extension). Maybe
+ * it's just execution ordering. (and I would guess it's core before ext)
+ */
+class HideTagsConfig
+{
+    public const ENABLED = "hide_tags_enabled";
+    public const VERSION = "hide_tags_version";
+}
+
+/**
  * Class Image
  *
  * An object representing an entry in the images table.
@@ -274,6 +287,8 @@ class Image
 
     private static function terms_to_conditions(array $terms): array
     {
+        global $config;
+
         $tag_conditions = [];
         $img_conditions = [];
         $stpen = 0;  // search term parse event number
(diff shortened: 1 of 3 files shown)


core/imageboard/image.php
```diff
@@ -292,6 +307,15 @@ private static function terms_to_conditions(array $terms): array
...
 
+        if ($config->get_bool(HideTagsConfig::ENABLED)) {
+            if (in_array('show_hidden', $terms)) {
+                $key = array_search('show_hidden', $terms);
+                unset($terms[$key]);
+            } else {
+                array_push($terms, '-hidden');
+            }
+        }
+
```
tegaki-tegaki, line 318: I wasn't entirely sure about this implementation... it's a bit strange to "inject" user query terms they never entered... but it _was_ a very easy way to create the feature I wanted 🤔

shish:
This seems like a useful feature - as to the implementation, adding a new search term into the query behind the scenes is probably the cleanest way; but right now it's only possible to add a whole extra condition to the database query, not possible to add a tag in among the other tags -- #920 should make this possible (and then in your own extension, you can react to `SearchTermParseEvent(term=null)` (ie, reacting to the fact that a search string is being parsed, not any specific term within the string) by adding a new `TagCondition(tag="hidden", positive=false)`)

PR #920, opened 2023-05-25
Allow SearchTermParseEvent to have a bit more control over results

shish:
Rather than "add querylet or do nothing", moving more code into the event means that event handlers are able to add a positive or negative querylet, add a positive or negative tag, or do nothing

This means that events can respond to the `null` search term by adding a tag, which would be useful for #917


shish:
(Merged manually)

PR #921, opened 2023-06-14
Approval bypass

luanalatte:
Hi! I implemented this quick change for images to be auto-approved on upload if the user has a special new permission to bypass the aproval process. A few things to note:

1. Possible improvement: I found it weird that I had to use the `approve_image` method instead of `$event->image->approved = true`, as this wasn't working. Is it because the `approved` boolean is not part of the default Image class?
2. I could've reused the existing permission for the bypass. After all, if you can approve posts you can approve your own. But doing it this way lets you assign only the bypass permission to a userclass, trusting all their posts by default but not letting them approve others'.
3. To keep the existing behaviour, I made a configuration option so that bypass is disabled by default. It can of course be removed if deemed unnecesary. This can also be achieved by simply removing the bypass permission from the admin userclass, but that would obscure the bypass functionality behind the need of a custom permissions file.

Let me know if this needs any changes :)

### DIFF
--- a/core/permissions.php
+++ b/core/permissions.php
@@ -110,6 +110,7 @@ abstract class Permissions
     public const CRON_ADMIN = "cron_admin";
     public const APPROVE_IMAGE = "approve_image";
     public const APPROVE_COMMENT = "approve_comment";
+    public const BYPASS_IMAGE_APPROVAL = "bypass_image_approval";
 
     public const SET_PRIVATE_IMAGE = "set_private_image";
     public const SET_OTHERS_PRIVATE_IMAGES = "set_others_private_images";
--- a/core/userclass.php
+++ b/core/userclass.php
@@ -217,6 +217,7 @@ new UserClass("admin", "base", [
 
     Permissions::APPROVE_IMAGE => true,
     Permissions::APPROVE_COMMENT => true,
+    Permissions::BYPASS_IMAGE_APPROVAL => true,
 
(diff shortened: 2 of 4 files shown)


shish:
Sorry for the delay, life is being life ^^;;


> To keep the existing behaviour, I made a configuration option so that bypass is disabled by default

I'm not sure yet, but my gut is telling me that the best thing would be to have this based only on user class, with admins being able to bypass by default -- and then to take care of visibility, we should really make the whole user class system more visible (I'm thinking at least in a read-only way, have a table of user classes on one axis and which permissions they have on the other, maybe with some way to add extra descriptions to each permission to describe exactly what it does 🤔  )

luanalatte:
> I'm not sure yet, but my gut is telling me that the best thing would be to have this based only on user class, with admins being able to bypass by default

Should I then simply remove the config option? Now that we have a simple way of letting admins know what permission is causing the auto-approval

shish:
> Should I then simply remove the config option?

Yeah, the more I'm thinking about this the more I like that direction :3

(Editing UserClass permissions is still a little fiddly because it requires editing files rather than point & click, but it feels like a fundamentally better approach - and now that I'm thinking about it, I wonder if we have any other board-level config options which would actually make more sense as user-level permissions...)

luanalatte:
Config has been removed :) admins will bypass approval by default.

shish:
Looks good, thank you :D

PR #922, opened 2023-06-25
update authors

luanalatte:
I... changed my name :P

I hope it's not a problem that I'm changing these old author lines. This is the same github account I used to originally push those, but if needed I can email you from the old address to prove it's me.

Thank you!


shish:
You seem trustworthy enough to me, and changing an author line shouldn't break anything~

PR #923, opened 2023-06-25
Make use of INSERT ... RETURNING

shish:
This is cleaner... but mysql ;(

### DIFF
--- a/core/database.php
+++ b/core/database.php
@@ -281,20 +281,6 @@ class Database
     }
 
     /**
-     * Get the ID of the last inserted row.
-     */
-    public function get_last_insert_id(string $seq): int
-    {
-        if ($this->get_engine()->id == DatabaseDriverID::PGSQL) {
-            $id = $this->db->lastInsertId($seq);
-        } else {
-            $id = $this->db->lastInsertId();
-        }
-        assert(is_numeric($id));
-        return (int)$id;
-    }
-
-    /**
      * Create a table from pseudo-SQL.
      */
     public function create_table(string $name, string $data): void
--- a/core/imageboard/image.php
+++ b/core/imageboard/image.php
@@ -360,7 +360,7 @@ class Image
         }
 
         if (is_null($this->id)) {
-            $database->execute(
+            $this->id = $database->get_one(
                 "INSERT INTO images(
 					owner_id, owner_ip,
                     filename, filesize,
@@ -374,7 +374,8 @@ class Image
 					:hash, :mime, :ext,
 				    0, 0,
 				    :posted, :source
-				)",
+				)
+                RETURNING id",
                 [
                     "owner_id" => $user->id, "owner_ip" => get_real_ip(),
                     "filename" => $cut_name, "filesize" => $this->filesize,
@@ -383,7 +384,6 @@ class Image
                     "posted" => $this->posted, "source" => $this->source
                 ]
             );
-            $this->id = $database->get_last_insert_id('images_id_seq');
         } else {
             $database->execute(
                 "UPDATE images SET ".
--- a/core/imageboard/tag.php
+++ b/core/imageboard/tag.php
@@ -105,12 +105,8 @@ class Tag
         );
         if (empty($id)) {
             // a new tag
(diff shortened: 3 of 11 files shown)


discomrade:
Is MySQL worth supporting? I haven't been able to find any advantage it has over MariaDB, except for proprietary customers trying to avoid a copyleft license (not a concern here). Its at the point where most common server Linux distributions just install mariadb when told to install mysql.

shish:
That is tempting...

The main reason to support it is that Shimmie was originally intended to be "the imageboard which anyone can install anywhere" (because ~20 years ago I was in high school and couldn't make Danbooru run on my $1/month shared LAMP web host). Today's entry-level webhosts (dreamhost, hostgator, etc) _still_ seem to be offering the original mysql 😭 

Buuuuuut... the entry-level webhosts do appear to have finally started offering vaguely-recent versions of PHP, which include SQLite, and IMO SQLite is a better choice for small-medium sites anyway, so I guess the question is how many users will I be screwing over if mysql support gets dropped now?

Looking at self-reported installs since we started collecting database version numbers in mid 2020:

```
     26    "mariadb"
     22    "mysql"
     16    "pgsql"
     23    "sqlite"
```

```
      1    "mariadb 10.1"
      6    "mariadb 10.3"
      4    "mariadb 10.4"
     10    "mariadb 10.5"
      3    "mariadb 10.6"
      1    "mariadb 10.8"
      1    "mariadb 11.2"
      8    "mysql 5.7"
     14    "mysql 8.0"
      1    "pgsql 10"
      3    "pgsql 11"
      2    "pgsql 12"
      5    "pgsql 13"
      4    "pgsql 14"
      1    "pgsql 16"
      2    "sqlite 3.26"
      2    "sqlite 3.27"
      1    "sqlite 3.28"
      1    "sqlite 3.30"
      2    "sqlite 3.31"
      1    "sqlite 3.33"
      5    "sqlite 3.34"
      4    "sqlite 3.36"
      1    "sqlite 3.37"
      1    "sqlite 3.39"
      3    "sqlite 3.40"
```

But then only looking at 2022 onwards:

```
     11    "mariadb"
      3    "mysql"
     10    "pgsql"
     15    "sqlite"
```

```
      1    "mariadb 10.4"
      5    "mariadb 10.5"
      3    "mariadb 10.6"
      1    "mariadb 10.8"
      1    "mariadb 11.2"
      2    "mysql 5.7"
      1    "mysql 8.0"
      1    "pgsql 11"
      5    "pgsql 13"
      3    "pgsql 14"
      1    "pgsql 16"
      1    "sqlite 3.27"
      1    "sqlite 3.31"
      5    "sqlite 3.34"
      3    "sqlite 3.36"
      1    "sqlite 3.37"
      1    "sqlite 3.39"
      3    "sqlite 3.40"
```

So of the three people who have checked-in recently and are using mysql, two are already using an explicitly-not-supported version 🤔 

Caveat that all the data collection is strictly opt-in, and most people who send a report do so one time straight after installation and never again, so I have no idea how representative it is (I don't even know where this sudden burst of bug reports and pull requests has come from after a few years of near-zero activity 😅 )

discomrade:
I wasn't thinking from a shared hosting perspective, you have a good point about staying accessible to hoster-provided LAMP offerings. Admins don't have much choice there.

>so I guess the question is how many users will I be screwing over if mysql support gets dropped now?

I wonder if adding a message to the next release/s asking for feedback on phasing out MySQL support will be insightful.

Are there any other improvements that could be made if MySQL support were dropped, or is it just this minor change? Because as much as I like code simplification and closing issues, it wouldn't be worth it just for this if it ruins any active boorus.

>I don't even know where this sudden burst of bug reports and pull requests has come from after a few years of near-zero activity 😅

For me personally, I've recently decided to migrate a booru.org site over to self-hosting, so I've been working on a few Gelbooru 0.1 features to make the change smoother. It wasn't due to any specific change with booru.org so I don't know where other contributors are coming from, but that's the great thing about an accessible tool that anyone can host - people can join and contribute when least expected :)

Issue #914, opened 2023-05-24
Please allow removing the post counter from homepage


diamondsw:
**Server Software**
```
about:
  title: "HomeBooru"
  theme: "danbooru2"

versions:
  shimmie: "2.10.0-alpha"
  schema: 21
  php: "8.2.5"
  db: "pgsql PostgreSQL 11.20 on x86_64-pc-linux-musl, compiled by gcc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924, 64-bit"
  os: "Linux 4d65c90ec21a 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21) x86_64"
  server: "PHP 8.2.5 Development Server"

extensions:
  core: ["admin","alias_editor","bbcode","comment","download","et","ext_manager","four_oh_four","handle_pixel","help_pages","image","index","media","mime","setup","static_files","system","tag_edit","tag_list","upgrade","upload","user","user_config","view"]
  extra: ["favorites","home","pools","random_image","random_list","rating","regen_thumb","relationships","tag_categories"]
  handled_mimes: ["image/jpeg","image/gif","image/png","image/webp"]

stats:
  images: 0
  comments: 0
  users: 2

media:
  memory_limit: "8.0MB"
  disk_use: "131.6GB"
  disk_total: "196.7GB"
```

 - Database: postgress
 - Web server: built into Docker container (add to documentation!)

**Client Software (please complete the following information)**
N/A

**What steps trigger this bug**
1. Go to "https://mybooruaddress/home" with the "home" extension enabled.
2. Observe that there is a post counter on the page that you'd really rather not see
3. Look in "Board Config" and see that there are only two options for the counter - "Default" and "newcounter", but not "None".

**What did you expect to happen?**
I hoped for a "None" option so I can remove the post counter from the home page entirely.

shish:
#924 took care of this :)

diamondsw:
Beautiful - thanks!

PR #924, opened 2023-06-26
option to hide post counters (home)

luanalatte:
Quick change addressing #914, could be improved but gets the job done.


shish:
Looks good to me, thank you ^_^

Issue #624, opened 2017-07-11
Show blocks only to specific user classes


PronMaster:
Hello is their anyway to limit the blocks extension for showing for different user classes, say like I want to show the anon user class the blocks and not show any for the admin user class is their any command I can use in the userclass.class.php to set this up or the ext blocks?

shish:
There's currently no way of doing that, and I can't think of a nice way to make that work :(

Leaving open as a feature request in case somebody can think of a good way to do it

PronMaster:
I hope this allowed but I'm willing to pay someone to develop a solution for this. This could be how pixiv do it and allow the site to have a premium service for supporters of the site which could help the site grow and reduce needed advertisement anyone interested?

MetallicAchu:
Cloud you please elaborate on what you mean by blocks? The news / ads stuff?

PR #926, opened 2023-06-26
show blocks based on user class

luanalatte:
Addressing #624, new documentation/help for the blocks extension might be needed.

Basically:

The userclass parameter can be left empty for blocks to show to everyone, or specified as a comma-separated, case-insensitive list of user classes that will see that block. Spaces around the comma get stripped.

Example, I have 3 blocks with the following content and userclass values:

- "You are not logged in": `anonymous`
- "You are logged in": `user, admin`
- "You are an admin": `admin`

A real use case:

- "<**ad here**>": `user, anonymous`
- "Thanks for supporting us!": `supporter`

### DIFF
--- a/ext/blocks/main.php
+++ b/ext/blocks/main.php
@@ -19,10 +19,15 @@ class Blocks extends Extension
 				title VARCHAR(128) NOT NULL,
 				area VARCHAR(16) NOT NULL,
 				priority INTEGER NOT NULL,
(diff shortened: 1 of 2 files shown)


shish:
> "\<ad here>": user, anonymous
> "Thanks for supporting us!": supporter

That's a really nice idea actually :O And the code looks good~

PR #941, opened 2023-07-05
More MicroHTML

luanalatte:
Turned `build_selector` into `SHM_SELECT` and realized `make_form_microhtml` was almost the same as `SHM_FORM`.

Also implemented a lot of MicroHTML changes on both the ratings and pools extensions. In this PR, I will try to do all of the functions in those two extensions as a starting point.

Merging isn't a hurry, unless you want those specific `SHM_` functions in master asap. For that reason, I'm creating the PR as a draft.


luanalatte:
Last commit has been a bit painful, and there's definitely room for improvement in how we do things, but at least we're not concatenating HTML tags anymore.

luanalatte:
I made minor microhtml changes to a couple other extensions. Plus the change in the index ext help. Not continuing that path as it would be out of this PR's scope.

Both rating's and pools' theme.php files are fully microhtml so this PR is ready, unless you want me to add more things to it.

shish:
Gah at my notifications not showing this x__x But it looks awesome, so merging now, thank you <3

PR #948, opened 2023-08-16
Use SCSS for CSS concatenation

shish:
SCSS
- does a better job of concatenation
- handles URLs better than a regex replacement
- generates a source map
- gives access to useful CSS functions like `lighter()` and `darker()` (if one renames the .css to .scss)


shish:
#956 does similar, better

PR #960, opened 2023-08-19
Add reverse image search links extension

thisisjoesGH:
Quality of life extension for taggers.

IQDB doesn't support images in WEBP format so I didn't include it in the supported services, but it could be added with a note in the extension documentation.

What it looks like:
Screenshot_20230819_152052

Tooltips on hover:
Screenshot_20230819_152225


shish:
That looks pretty handy~ I am 50/50 on having the links in their own block vs putting them into the ImageInfoBlock (so that they'd be right next to the tag input field) - not sure if that's actually better or worse, just the first thing that came to mind ^^

Either way it seems like a fairly straightfoward and useful extension, so I'll try to merge tomorrow (It's bed time now and I'm too sleepy to code-review safely ^^;; )

thisisjoesGH:
The default position and priority should place the block right below the image info (See https://booru.oke.moe/post/view/671 for an example). I figured some admins might like to place it somewhere else on the page like in the left navbar.

I could always add an onImageInfoBoxBuilding event handler and an option to stick the reverse search links in the ImageInfo block.

shish:
Thinking some more, my gut says that "where information is displayed" should be left to the theme, rather than each extension having its own "where should I appear?" options on the setup screen - so I'd do something along the lines of:

`ext/reverse_image_search/main.php`:
```
function onImageInfoBlockBuilding($event) {
    // Using onImageInfoBlockBuilding instead of onDisplayingImage means
    // that we have the choice of either adding ourselves to the info block
    // or somewhere else
    $this->theme->display_links($event);
}
```

`ext/reverse_image_search/theme.php`:
```
function display_links($event) {
    // display the links in their own separate block
    $page->add_block(new Block("Reverse Image Search", $this->generate_links($event->image), "main", 20));
}
```

And then if somebody wants to have it in the info block, they can override display_links like

`themes/my_custom_theme/reverse_image_search.theme.php`:
```
function display_links($event) {
    // add links into the image info block
    $event->add_part([...], $this->generate_links($event->image), [...]);
}
```

For each individual extension, using theme overrides is a little more effort than selecting from a drop-down menu -- but then theme overrides are a standard approach that works for all extensions, so I think overall it ends up being simpler than each extension using its own system 🤔

(I'd be happy to merge and make these changes myself, I mostly want to check that that won't conflict with anything you have planned ^^)

shish:
(Also now that I think about it, I wonder if `DisplayingImageEvent` and `ImageInfoBoxBuildingEvent` should actually be merged... but that's something for me to think about later rather than something that needs solving now ^^)

shish:
Merged manually, minus the customisability for now, because I want to think harder about a more standard layout-customisation system instead of each extension having its own ^^ Thank you <3

Issue #904, opened 2023-03-15
New CVE-2023-27522 Vulnerability and "%20" space encoding.


JamesTheSergal:
**Some very quick background -**
I just updated my web host with the newest version of Apache2 for security, and stability.
A vast majority of people won't be affected by this yet.

**Server Software**
about:
  title: "[Removed]"
  theme: "danbooru2"
  url: "[Removed]"

versions:
  shimmie: "2.9.1" (Only slightly modified)
  schema: 21
  php: "8.1.2-1ubuntu2.11"
  db: "mysql 10.6.12-MariaDB-0ubuntu0.22.04.1"
  os: "Linux 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64"
  server: "Apache/2.4.52 (Ubuntu)"

extensions:
  core: ["admin","alias_editor","bbcode","comment","download","et","ext_manager","four_oh_four","handle_pixel","help_pages","image","index","media","mime","setup","static_files","system","tag_edit","tag_list","upgrade","upload","user","user_config","view"]
  extra: ["auto_tagger","autocomplete","biography","blocks","blotter","browser_search","bulk_actions","bulk_download","custom_html_headers","danbooru_api","downtime","favorites","handle_svg","handle_video","home","image_hash_ban","image_view_counter","link_image","numeric_score","ouroboros_api","pm","pools","post_peek","random_image","random_list","rating","regen_thumb","relationships","report_image","shimmie_api","site_description","source_history","tag_categories","tag_editcloud","tag_history","tag_tools"]
  handled_mimes: ["image/jpeg","image/gif","image/png","image/webp","image/svg+xml","video/x-ms-asf","video/x-msvideo","video/x-flv","video/x-matroska","video/mp4","video/ogg","video/quicktime","video/webm"]

stats:
  images: 151447
  comments: [Removed]
  users: [Removed]

media:
  memory_limit: "64.0MB"
  disk_use: "380.9GB"
  disk_total: "491.1GB"

thumbnails:
  engine: "convert"
  quality: 100
  width: 128
  height: 128
  scaling: 400
  mime: "image/jpeg"

 - Shimmie version: 2.9.1 (Only slightly modified)
 - Database: mysql
 - Web server: Apache2 2.4.52

**Client Software**
 - Device: Windows 11
 - Browser: Chrome

**What steps trigger this bug**
[...]

PR #964, opened 2023-09-30
Fix .htaccess rewrite

intergalacticmonkey:
When using recent versions of Apache and the provided `.htaccess` file (with "Nice URLs" enabled in Shimmie), you get 403 Forbidden when attempting to navigate to a URL containing spaces, e.g. `http://<shimmie>/post/list/tag1%20tag2/1`.

Apache's error log shows:
```
AH10411: Rewritten query string contains control characters or spaces
```

We fix this by adding `B` and `BNP`, as per [here](https://webmasters.stackexchange.com/a/141978).

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -21,7 +21,7 @@
 
 	# any requests for files which don't physically exist should be handled by index.php
 	RewriteCond %{REQUEST_FILENAME} !-f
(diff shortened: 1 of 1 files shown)


JamesTheSergal:
Relates to #904

shish:
I don't have an apache server to test with, but this seems sensible from the documentation, thank you <3

PR #970, opened 2023-11-08
Autocomplete and search for alias_editor and auto_tagger

yls4:
Enable autocomplete for adding new tags
Enable search function

### DIFF
--- a/ext/alias_editor/main.php
+++ b/ext/alias_editor/main.php
@@ -37,8 +37,8 @@ class AliasTable extends Table
         $this->size = 100;
         $this->limit = 1000000;
         $this->set_columns([
-            new TextColumn("oldtag", "Old Tag"),
-            new TextColumn("newtag", "New Tag"),
+            new AutoCompleteColumn("oldtag", "Old Tag"),
+            new AutoCompleteColumn("newtag", "New Tag"),
             new ActionColumn("oldtag"),
         ]);
         $this->order_by = ["oldtag"];
--- a/ext/alias_editor/main.php
+++ b/ext/alias_editor/main.php
@@ -61,8 +61,8 @@ class AliasTable extends Table
         }
         $thead->appendChild($tr);
 		
-		if ($this->create_url) {
-            $tr = TR();
+		$tr = TR();
+		if ($this->search_url) {
             foreach ($this->columns as $col) {
                 $tr->appendChild(TD($col->read_input($this->inputs)));
             }
--- a/ext/alias_editor/main.php
+++ b/ext/alias_editor/main.php
@@ -8,8 +8,26 @@ use MicroCRUD\ActionColumn;
 use MicroCRUD\TextColumn;
 use MicroCRUD\Table;
 
+use MicroHTML\HTMLElement;
+
+use function MicroHTML\emptyHTML;
+use function MicroHTML\TABLE as html_TABLE;
+use function MicroHTML\THEAD;
+use function MicroHTML\TBODY;
+use function MicroHTML\TFOOT;
+use function MicroHTML\TR;
+use function MicroHTML\TH;
+use function MicroHTML\TD;
+use function MicroHTML\INPUT;
+use function MicroHTML\FORM;
+use function MicroHTML\DIV;
+use function MicroHTML\A;
+use function MicroHTML\B;
+use function MicroHTML\BR;
+
 class AliasTable extends Table
 {
+	public $search_url = null;
     public function __construct(\FFSPHP\PDO $db)
     {
         parent::__construct($db);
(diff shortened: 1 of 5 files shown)


ext/alias_editor/main.php
```diff
@@ -91,7 +136,17 @@ public function onPageRequest(PageRequestEvent $event)
...
                 $t->size = $config->get_int('alias_items_per_page', 30);
+				$input = validate_input(["r_oldtag"=>"string,optional", "r_newtag"=>"string,optional"]);
+				$tag_inputs = array();
+				if (isset($_GET["r_oldtag"])) {
+					$tag_inputs["r_oldtag"] = $_GET["r_oldtag"];
+				}
+				if (isset($_GET["r_newtag"])) {
+					$tag_inputs["r_newtag"] = $_GET["r_newtag"];
+				}
+				$t->inputs = array_merge($t->inputs, $input, $tag_inputs);
```
shish, line 147: This looks like a complicated way of taking inputs from both GET and POST at the same time? But it does so using 3 different mechanisms (and it's not clear why anything beyond the default GET support is needed). If we do actually need to support both GET and POST I'd do it with `$t->inputs = array_merge($_GET, $_POST)`

ext/alias_editor/theme.php
```diff
@@ -30,7 +30,11 @@ public function display_aliases(HTMLElement $table, HTMLElement $paginator): voi
         $page->set_title("Alias List");
         $page->set_heading("Alias List");
         $page->add_block(new NavBlock());
-        $page->add_block(new Block("Aliases", $html));
+		$block = new Block("Aliases", $html);
+		$block->body = str_replace(array("name='c_oldtag'", "name='c_newtag'"), 
+						array("name='c_oldtag' class='autocomplete_tags'", "name='c_newtag' class='autocomplete_tags'"), 
+						$block->body);
+        $page->add_block($block);
```
shish, line 37: Doing search & replace on generated HTML is very fragile - I'm guessing that the intention here is to add `class='autocomplete_tags'` to the "add alias" inputs? I'd do that by going to the `AliasTable` definition and replacing `TextColumn` with `TagColumn` (a new class which subclasses `TextColumn` and overrides the `create_input()` method)

ext/alias_editor/main.php
```diff
@@ -91,7 +136,17 @@ public function onPageRequest(PageRequestEvent $event)
...
                 if ($user->can(Permissions::MANAGE_ALIAS_LIST)) {
+					$t->search_url = make_link("alias/list");
```
shish, line 149: The default target for an HTML form is "whatever page the form is on", which seems correct when we are adding search parameters to the current page - why override that?

shish (changes requested):

ext/alias_editor/main.php
```diff
@@ -26,6 +44,33 @@ public function __construct(\FFSPHP\PDO $db)
...
+            }
+        }
+        $thead->appendChild($tr);
+		
+		if ($this->create_url) {
+            $tr = TR();
+            foreach ($this->columns as $col) {
+                $tr->appendChild(TD($col->read_input($this->inputs)));
+            }
+        }
```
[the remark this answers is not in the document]
yls4: Most of this code was copypasted from the class it's inheriting, probably a bad copypaste job.

ext/alias_editor/main.php
yls4: The search button and the pagination uses different mechanisms.  If I don't do it this way, then you can search for a tag but if you try to go to page 2, the search fields will not persist.

ext/alias_editor/main.php
yls4: When I was testing this without that it took me to the home page instead

ext/auto_tagger/main.php
```diff
@@ -245,6 +300,38 @@ private function add_auto_tag(string $tag, string $additional_tags)
                 AutoTaggerInfo::KEY,
                 "Added auto-tag for {$tag} -> {".implode(" ", $additional_tags)."}"
             );
+        }
+		// Now we apply it to existing items
+		$this->apply_new_auto_tag($tag);
+    }
+
+    private function update_auto_tag(string $tag, string $additional_tags): bool
```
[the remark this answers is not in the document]
yls4: I think that's was part of an old change I made a long time ago to make updates to existing tags concatenate instead of replace.  I don't see it called anywhere in my own branch either so I must have decided to do it another way.

ext/alias_editor/main.php
shish: Ooooh do you not have niceurls enabled? Now that you mention it I can see how this would be broken in that case, I think it should be possible to come up with a more general upstream fix that handles all of the MicroCRUD tables at once, lemme try and do that /o/

shish:
I've bumped all the dependency versions and pulled in a new version of MicroCRUD which should make searching work both with and without niceurls :3

ext/alias_editor/theme.php
yls4: Which directory would you like me to define a TagColumn class?

shish: That is a good question - I would put it in `core/imageboard/tagcolumn.php` (`core` because it is something that seems generally useful for multiple extensions, and `imageboard` because that's where tag-management code currently lives)

shish:
This got merged manually ^^

PR #1031, opened 2024-02-12
[autocomplete] support aliases

shish:


shish:
merged by accident >.>

PR #1035, opened 2024-02-15
Add togglable thumbnails for post relationships

thisisjoesGH:
Can show thumbnails for parents, children, and siblings! should not show any relationships if they are hidden by ratings/etc.

A lot of shimmie code has been refactored since this was originally written, so I've done my best to adapt it. if you have suggestions for optimizing the code or making it more readable lmk

**Child post thumbnails:**
2024-02-14 22 34 12 localhost 5eb54811cad2

**Parent and sibling thumbnails:**
2024-02-14 22 34 17 localhost fe84d3cf9397

### DIFF
--- a/ext/relationships/main.php
+++ b/ext/relationships/main.php
@@ -172,18 +172,14 @@ class Relationships extends Extension
     /**
      * @return Image[]
      */
-    public static function get_children(Image $image, int $omit = null): array
+    public static function get_children(int $image_id): array
     {
         global $database;
-        $results = $database->get_all_iterable("SELECT * FROM images WHERE parent_id = :pid ", ["pid" => $image->id]);
-        $output = [];
-        foreach ($results as $result) {
-            if ($result["id"] == $omit) {
-                continue;
-            }
-            $output[] = new Image($result);
-        }
-        return $output;
+        $child_ids = $database->get_col("SELECT id FROM images WHERE parent_id = :pid ", ["pid" => $image_id]);
+
+        $children = Search::get_images($child_ids);
+
+        return $children;
     }
 
     private function remove_parent(int $imageID): void
@@ -216,5 +212,37 @@ class Relationships extends Extension
             "UPDATE images SET has_children = :has_children WHERE id = :pid",
             ["has_children" => $children > 0, "pid" => $parent_id]
         );
+    }
+
+    public static function has_siblings(int $image_id): bool
+    {
+        global $database;
+
+        $image = Image::by_id($image_id);
+
+        $count = $database->get_one(
+            "SELECT COUNT(*) FROM images WHERE id!=:id AND parent_id=:pid",
+            ["id" => $image_id, "pid" => $image['parent_id']]
+        );
+
+        if ($count > 0) {
+            return true;
+        }
+        return false;
+    }
+
+    public static function get_siblings(int $image_id): array
+    {
+        global $database;
+
+        $image = Image::by_id($image_id);
+
+        $sibling_ids = $database->get_col(
+            "SELECT id FROM images WHERE id!=:id AND parent_id=:pid",
+            ["id" => $image_id, "pid" => $image['parent_id']]
+        );
+        $siblings = Search::get_images($sibling_ids);
+
+        return $siblings;
     }
 }
--- a/ext/relationships/theme.php
+++ b/ext/relationships/theme.php
@@ -14,22 +14,48 @@ class RelationshipsTheme extends Themelet
     {
         global $page, $database;
 
-        if ($image['parent_id'] !== null) {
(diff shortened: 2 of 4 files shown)


thisisjoesGH:
Seems like PHPStan doesn't like the nested array returned by get_siblings. not sure what the best way to make it happy is

shish:
phpstan wants arrays to be more specific than "array", because in PHP an array can be a list or a map, with strings or numbers (or both) as keys, and anything as values -- the fix is to add a PHPDoc comment to the function like

```
/**
 * @return Image[]
 */
```

(Looking through the rest now /o/)

ext/relationships/main.php
```diff
@@ -172,18 +172,14 @@ public function onImageRelationshipSet(ImageRelationshipSetEvent $event): void
...
-                continue;
-            }
-            $output[] = new Image($result);
-        }
-        return $output;
+        $child_ids = $database->get_col("SELECT id FROM images WHERE parent_id = :pid ", ["pid" => $image_id]);
+
+        $children = Search::get_images($child_ids);
+
+        return $children;
```
shish, line 182:
```suggestion
        return Search::get_images($child_ids);
```

ext/relationships/main.php
```diff
@@ -217,4 +213,36 @@ private function set_has_children(int $parent_id): void
...
+
+        $count = $database->get_one(
+            "SELECT COUNT(*) FROM images WHERE id!=:id AND parent_id=:pid",
+            ["id" => $image_id, "pid" => $image['parent_id']]
+        );
+
+        if ($count > 0) {
+            return true;
+        }
+        return false;
```
shish, line 231:
```suggestion
        return $count > 0;
```

ext/relationships/theme.php
```diff
@@ -14,22 +14,48 @@ public function relationship_info(Image $image): void
...
-            foreach ($ids as $id) {
-                $html .= "#<a href='".make_link('post/view/'.$id)."'>{$id}</a>, ";
+            $visible_children = Relationships::get_children($image->id);
+            if (!empty($visible_children)) {
+                $child_summary_html = "<span>This post has <a href='".make_link('post/list/parent='.$image->id.'/1')."'>".(count($visible_children) > 1 ? "child posts" : "a child post")."</a>";
+                $child_summary_html .= " (post ";
+                $child_thumb_html = "<div class='shm-relationships-child-thumbs'><div class='shm-child-thumbs'>";
+                foreach ($visible_children as $child) {
+                    $child_summary_html .= "<a href='".make_link('post/view/'.$child->id)."'>#{$child->id}</a>, ";
+                    $child_thumb_html .= $this->get_child_thumbnail_html(Image::by_id($child->id));
```
shish, line 51: Isn't `$child` already an `Image` object? o.o
```suggestion
                    $child_thumb_html .= $this->get_child_thumbnail_html($child);
```

ext/relationships/theme.php
```diff
@@ -70,4 +96,37 @@ public function get_help_html(): string
...
+    private function get_parent_thumbnail_html(Image $image): HTMLElement
+    {
+        global $user;
+
+        $parent_id = $image['parent_id'];
+        $parent_image = Image::by_id($parent_id);
+
+        $html = $this->build_thumb_html($parent_image);
+
+        return $html;
```
shish, line 109:
```suggestion
        $parent_id = $image['parent_id'];
        $parent_image = Image::by_id($parent_id);

        return $this->build_thumb_html($parent_image);
```

ext/relationships/theme.php
```diff
@@ -70,4 +96,37 @@ public function get_help_html(): string
...
+    {
+        $html = $this->build_thumb_html($image);
+
+        return $html;
+    }
+
+    private function get_sibling_thumbnail_html(Image $image): string
+    {
+        global $user;
+
```
shish, line 122:
```suggestion
```

ext/relationships/main.php
```diff
@@ -217,4 +213,36 @@ private function set_has_children(int $parent_id): void
...
+    }
+
+    public static function get_siblings(int $image_id): array
```
shish, line 234: To indicate that this function returns "an array of Images", rather than "an array of anything"
```suggestion
    /**
     * @return Image[]
     */
    public static function get_siblings(int $image_id): array
```

shish (changes requested):
Looking good /o/ The only required fix is the type hint on `get_siblings`, the others are minor :)

ext/relationships/main.php
thisisjoesGH: fixed

ext/relationships/theme.php
thisisjoesGH: oop, missed that when adapting the old code. fixed

shish:
All looks great, merged now, thank you :)

PR #1038, opened 2024-02-16
Do not cast hard drive size to integer - fixes #1037

### LINKED ISSUE #1037: System info page crashes when running on a 32-bit OS
**Server Software**
(You can get all these stats from `http://<your site>/system_info`)

 - Shimmie version: 2.10.6
 - Database: sqlite
 - Web server: nginx + php-fpm

 - Device: irrelevant
 - Browser: irrelevant

1. Install on a 32-bit machine, such as a Raspberry Pi
2. Click on system info

The system information is shown.

**What actually happened?**
A blank screen was presented, due to a integer overflow when calculating the total disk usage and total space.

Function `get_info()` in `/ext/et/main.php` casts the output of `disk_total_space` and `disk_free_space` to an integer. This integer is platform-dependant. In 32-bit machines like most embedded ARM devices or SBCs, this integer is 32-bit so it overflows when the disk is larger than 2GB.

This causes the application to crash when entering said page:
```
2024/02/16 16:25:02 [error] 425#425: *85 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: Shimmie2\to_shorthand_int(): Argument #1 ($int) must be of type int, float given, called in /mnt/data/shimmie2/ext/et/main.php on line 108 and defined in /mnt/data/shimmie2/core/polyfills.php:491
Stack trace:
#0 /mnt/data/shimmie2/ext/et/main.php(108): Shimmie2\to_shorthand_int()
#1 /mnt/data/shimmie2/ext/et/main.php(21): Shimmie2\ET->get_info()
#2 /mnt/data/shimmie2/core/send_event.php(161): Shimmie2\ET->onPageRequest()
#3 /mnt/data/shimmie2/index.php(91): Shimmie2\send_event()
#4 {main}
  thrown in /mnt/data/shimmie2/core/polyfills.php on line 491" while reading response header from upstream, client: 192.168.0.35, server: <ip>, request: "GET /shimmie/index.php?q=system_info HTTP/1.1", upstream: "fastcgi://unix:/run/php/php-fpm.sock:", host: "<ip>", referrer: "http://<ip>/shimmie/"
```

socram8888:
`disk_total_space` and `disk_free_space` return floats already, so we can avoid triggering an integer overflow by just working using those floats instead.

### DIFF
--- a/core/polyfills.php
+++ b/core/polyfills.php
@@ -488,7 +488,7 @@ function parse_shorthand_int(string $limit): int
 /**
  * Turn an integer into a human readable filesize, eg 1024 -> 1KB
  */
-function to_shorthand_int(int $int): string
+function to_shorthand_int(int|float $int): string
 {
     assert($int >= 0);
 
--- a/ext/et/main.php
+++ b/ext/et/main.php
@@ -77,6 +77,8 @@ class ET extends Extension
             $ver .= "+";
         }
 
+        $disk_total = false_throws(disk_total_space("./"));
+        $disk_free = false_throws(disk_free_space("./"));
         $info = [
             "about" => [
                 'title' => $config->get_string(SetupConfig::TITLE),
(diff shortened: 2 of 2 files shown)


shish:
Nice catch :D Looks like those functions return `float|false` though - you can use `false_throws(thing)` to turn `thing` from `X|false` into `X` (throwing an exception if it is false, which is better than the current "silently do the wrong thing and continue if there's an error")

socram8888:
I've just force-pushed a new version that does use false_throws. I've moved the disk_*_space to their own lines, so they're called only once (as disk_total_space was called twice), and it reduces the amount of code in a single line.

shish:
Excellent, thank you <3

PR #1039, opened 2024-02-16
Fix order:random not working on SQlite

socram8888:
Currently order:random_XXXX is broken on SQlite (and AFAIK PostgreSQL too) due to RAND() being called with a seed, which is not supported.

This pull requests adds a new call to the Database which generates SQL code that, depending on the engine, will keep using the fast seeded RAND(N) if supported, or fall back to using a DRBG based on MD5.

It has been added to Database instead of the Main extension so it can be reused elsewhere if needed.

### DIFF
--- a/core/database.php
+++ b/core/database.php
@@ -439,4 +439,19 @@ class Database
             $this->execute("ALTER TABLE $table RENAME COLUMN {$column}_b TO $column");
         }
     }
+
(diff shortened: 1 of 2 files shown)


shish:
Looks good to me, and I've added #1040 as a test for it :)

PR #1042, opened 2024-02-17
[upload] add rating fields to upload form

discomrade:
If the Post Ratings extension is enabled, encourage a rating to be selected when uploading.

This can be extended to also have a Common Ratings field, if wanted.

### DIFF
--- a/ext/rating/main.php
+++ b/ext/rating/main.php
@@ -165,6 +165,11 @@ class Ratings extends Extension
         }
     }
 
+    public function onUploadRatingBuilding(UploadRatingBuildingEvent $event): void
+    {
+        $event->part = (string)$this->theme->get_upload_rater_html($event->suffix);
+    }
+
     public function onDisplayingImage(DisplayingImageEvent $event): void
     {
         global $page;
@@ -204,7 +209,7 @@ class Ratings extends Extension
     {
         global $user;
         $event->add_part(
-            $this->theme->get_rater_html(
+            $this->theme->get_image_rater_html(
                 $event->image->id,
                 $event->image['rating'],
                 $user->can(Permissions::EDIT_IMAGE_RATING)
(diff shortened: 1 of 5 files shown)


shish:
I like the direction this is going, it's actually making me think about going even further, making _all_ of the upload fields optional and added at runtime by their separate extensions 🤔  I'm going to merge this as-is though, as it's one step forwards and I'd like to do another big refactor before getting to that ^^

shish:
Argh, actually reverting this - it looks like the unit test setup for some reason didn't have assertions enabled, so this is making a bunch of things fail but they fail silently >.< Lemme see if I can fix that and then run the tests again...

discomrade:
Sorry about that, I didn't notice failures when I tested.

Great to hear you're interested in taking it further with upload fields ^^ I added ratings first because it's an important field when it's enabled e.g. if explicit and safe are both allowed on the same booru and we're using CSS to hide/blur explicit thumbnails from sensitive users then it's important to make sure its tagged. I'm considering making it a mandatory field on my booru.

shish:
#1052 does a bunch of refactoring which makes upload and info-set share code so one doesn't need to implement separate code for both of them, and also makes the common case easier (and also stops "tags" and "source" from being special - they are now treated the same as any other post metadata).

So now, anything which is being set during ImageInfoSet based on `$_POST["foo"]` can now also be set during upload based on `$_POST["foo$N"]` or `$_POST["foo"]`

So... for ratings to be set during upload, the only changes necessary *should* be just adding either a common "rating" input, or a file-specific "rating$N" input, to the upload form.

discomrade:
For a non-core extension (like ratings), does `build_upload_list()` need to have build events to allow extensions to append their own input/select elements to the common and post-specific sections?

shish:
> does build_upload_list() need to have build events to allow extensions to append their own input/select elements to the common and post-specific sections

Yeah, I haven't thought of anything more specific, but that is the general direction I'm imagining :)

PR #1049, opened 2024-02-20
[notes] allow clicking through notes container

discomrade:
If the notes extension is enabled, the container would block clicking the image, including right-clicking to save it


shish:
Aha, that was bothering me too - thank you for figuring out how to fix it <3

PR #1051, opened 2024-02-20
Various permissions fixes for forum and notes extensions

discomrade:
Each commit has a description of its changes ^^

### DIFF
--- a/core/permissions.php
+++ b/core/permissions.php
@@ -125,6 +125,7 @@ abstract class Permissions
     public const NOTES_ADMIN = "notes_admin";
     public const NOTES_CREATE = "notes_create";
     public const NOTES_EDIT = "notes_edit";
+    public const NOTES_REQUEST = "notes_request";
 
     public const POOLS_ADMIN = "pools_admin";
     public const POOLS_CREATE = "pools_create";
--- a/core/permissions.php
+++ b/core/permissions.php
@@ -120,7 +120,7 @@ abstract class Permissions
     public const BYPASS_IMAGE_APPROVAL = "bypass_image_approval";
 
     public const FORUM_ADMIN = "forum_admin";
-    public const FORUM_CREATE_THREAD = "forum_create_thread";
+    public const FORUM_CREATE = "forum_create";
 
     public const NOTES_ADMIN = "notes_admin";
     public const NOTES_CREATE = "notes_create";
--- a/core/userclass.php
+++ b/core/userclass.php
@@ -119,6 +119,7 @@ new UserClass("user", "base", [
     Permissions::FORUM_CREATE => true,
     Permissions::NOTES_CREATE => true,
     Permissions::NOTES_EDIT => true,
+    Permissions::NOTES_REQUEST => true,
     Permissions::POOLS_CREATE => true,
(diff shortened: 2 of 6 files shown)


shish:
> make admin class a child of user 

one downside of that is that if somebody decides to customise the `user` class to remove a permission (eg "Only moderators and admins should be able to set post ratings"), they then *also* need to customise the `admin` class to re-add it - which isn't the end of the world, but could be confusing

Now that I'm thinking about this, I think the ideal thing would be for `admin` to have a dynamically generated permission list - like how `base` is automatically set to have all permissions `false`, `admin` should automatically be all-`true` (with an override for `Permissions::HELLBANNED => false`, because that's special) 🤔

shish:
The changes for forum and notes all look good, no comments there :)

discomrade:
> one downside of that is that if somebody decides to customise the `user` class to remove a permission (eg "Only moderators and admins should be able to set post ratings"), they then _also_ need to customise the `admin` class to re-add it - which isn't the end of the world, but could be confusing

It can be a possible trap if adjusting an existing permission to remove it from only the user, not realizing it needed to be added to its child role. Do you think having documentation on editing roles, either in code comments or on here, would mitigate this?

I also think this approach will be more useful if other roles are added, like a moderator (e.g. can delete/edit content, but not board settings), where an admin would become a child of a moderator.

> Now that I'm thinking about this, I think the ideal thing would be for `admin` to have a dynamically generated permission list - like how `base` is automatically set to have all permissions `false`, `admin` should automatically be all-`true` (with an override for `Permissions::HELLBANNED => false`, because that's special) 🤔

It's a cool idea, although I think dynamic permission lists could become confusing if another anti-permission like Permissions::HELLBANNED is added, especially if it's just being tested and suddenly the admin user is affected when it's not even (manually) added to them or the user.

shish:
> Do you think having documentation on editing roles, either in code comments or on here, would mitigate this?

A slight tangent, but at some point I'd love to have a GUI permission editor, which might work as a form of documentation so one can see the effects of enabling or disabling each flag 🤔 


> an admin would become a child of a moderator.

From my perspective, "admin" is defined as "has access to all current *and future* permissions", which feels fundamentally different to "a moderator + [some fixed list of extra things]"


> dynamic permission lists could become confusing if another anti-permission like Permissions::HELLBANNED is added

TBH I'd rather clear up that confusion by getting rid of the hellbanned permission - it isn't really a "permission" so much as "a weird special-case behaviour which applies to one particular class of user" (and even using the user-class system is a hack, since it's more like a boolean which might apply to any user regardless of class...)

shish:
I've manually pulled and merged the other two commits - it's just the change from "an admin is somebody with all permissions" to "an admin is a regular user plus some extra bits" that I'm not convinced by -- I'm not entirely against it, but my gut is telling me to lean harder into "an admin is somebody with all permissions" (ie #1055) 👀

discomrade:
Sounds good :) Another software I've used had a special `Root` role, different to the lesser `Admin` role, which is why I was thinking the way I was, but really there's no point separating it out like that. It just makes it confusing. Having an admin with all permissions makes sense.

Closing this merge request as completed by #1055

PR #1053, opened 2024-02-21
Installation Fixes and Default Config Changes

NottyNoz:
- core/install.php install form POST pointed to wrong location when the booru is in a subdirectory
- core/sanitize_php.php die_nicely is used by the installer and the static files point to the wrong location when the booru is in a subdirectory
- an email being required should not be the default since during installation the admin would need to enter one, there is an easy checkbox to enable it in the board config, and user emails aren't validated and generally unused except for gravatar
- changed the default zoom to be "both" since every other booru downsizes a post to fit it within view, nobody wants to just see just half of a large post by default

### DIFF
--- a/core/install.php
+++ b/core/install.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Shimmie2;
 
+require_once "core/urls.php";
+
 /**
  * Shimmie Installer
  *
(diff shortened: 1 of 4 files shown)


shish:
Yep that all seems sensible /o/

PR #1060, opened 2024-02-21
Style counter images on the home page to adjust to screen width

discomrade:
Normal: (unchanged)

image

Narrow: (previously hidden)

image


shish:
Looks good, the failing test is my fault, should be fixed in #1062

Issue #292, opened 2013-05-18
OR operator


shish:
Can it be done without overloading the database?

discomrade:
I am currently testing a basic OR operator: `~tag1 ~tag2 tag3` matches posts with tag3 and either tag1 and/or tag2 . `~tag1 ~tag2 -tag3` matches posts with either tag1 and/or tag2 , except any with tag3. Order doesn't matter, there can be any number of OR operations. It does not work with parentheses.

It also handles negated OR: `~tag1 ~-tag2 ~-tag3` matches posts either with tag1 or which don't contain tag2 or which don't contain tag3 (so a post with `tag2 tag4` will match because it doesn't have tag3, but a post with `tag2 tag3` will not match)

While I do think OR with parentheses is much easier to use and powerful, they're too advanced for my current experience with parsing and SQL. This format is common enough (used on e621ng, legacy operator on danbooru, documented but 'not working' on safebooru, even though another Gelbooru 0.2 site has parentheses working)

PR #1082, opened 2024-02-24
Optimisations for search queries

discomrade:
Disclaimer: I'm not experienced with SQL, so feel welcome to correct any mistaken assumptions I've made.

The first commit prevents tag queries from selecting all columns when only the id is being used, improving the speed of the join operations.
The second commit prevents all columns being fetched when only the total number of rows is relevant.

These are both improvements by themselves, although the first is also a useful step towards other optimisations, especially those needed to make INTERSECT and UNION operations efficient, which are necessary for https://github.com/shish/shimmie2/issues/292

### DIFF
--- a/core/imageboard/search.php
+++ b/core/imageboard/search.php
@@ -244,10 +244,11 @@ class Search
         ?int $limit = null,
         ?int $offset = null
     ): Querylet {
+        $query = new Querylet("SELECT images.* FROM images INNER JOIN (");
         // no tags, do a simple search
         if (count($tag_conditions) === 0) {
             static::$_search_path[] = "no_tags";
-            $query = new Querylet("SELECT images.* FROM images WHERE 1=1");
+            $query->append(new Querylet("SELECT images.id FROM images WHERE 1=1"));
         }
 
         // one tag sorted by ID - we can fetch this from the image_tags table,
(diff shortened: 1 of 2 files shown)


shish:
So I was looking at this and thinking "it's a great idea, _something_ feels more complicated than needed though" - after much thinking and failing to put my finger on it, I opened up my editor and tried to figure out how I would write this code if I were doing it from scratch. But before I'd actually managed to write anything, Copilot had auto-completed what appears to be an even simpler and faster approach than either of our ideas :O The result is at #1090


> a useful step towards other optimisations, especially those needed to make INTERSECT and UNION operations efficient, which are necessary for https://github.com/shish/shimmie2/issues/292

I wonder about this though - if your approach is a _little_ bit more complicated, but opens the door to useful features in a way that #1090 doesn't, then I would go with this approach 👀

discomrade:
I'm ready to upload my disjunctive ('OR') code as a draft pull request for reference, although be aware that I based it over the two commits in this merge request so I will need to force commit to it.

By using `images.id` instead of `images.*` until a final JOIN INNER, I'm able to significantly improve the speed, so I believe the co-pilot approach can be adapted but it shouldn't be considered a replacement for the `[core] only select id column on tag queries` commit in this merge request until it uses `.id`.

Using .id only also allows me to avoid using INTERSECT, in favour of the much faster AND [x] IN ([y]), when factoring in disjunctive results. This reduces the query time by orders of magnitude, so I consider it necessary.

discomrade:
Marked as draft to avoid this being merged as-is:
- If #1090 is merged, this will need to be resubmitted.
- If #1153 is merged, this will need to be refactored.

This code should be rewritten and resubmitted after those pull requests are accepted.

PR #1090, opened 2024-02-24
[core] simplify counting search results

shish:
[core] simplify counting search results

`SELECT images.*` vs `SELECT COUNT(*)`


shish:
(Also TIL that `COUNT(*)` is literally interpreted as "count the number of rows" and it doesn't load all of the columns from disk despite what the `*` might imply, it can actually be _faster_ than `COUNT(id)`)

discomrade:
This will need to be re-written for it to work with #1153

PR #1092, opened 2024-02-24
Add disjunctive search operation

discomrade:
**Work in progress - do not merge**

This commit adds the disjunctive operands to search, e.g. `~computer ~thing ~score:>5` (means: computer OR thing OR score:>5), as well as negative disjunctive operands e.g. `~cat ~-phone` (means: cat OR NOT phone). This also works with image conditions.


shish:
(Sorry for the delays here - this seems pretty great but I have some foundational bits I'd like to merge first and things at my day job got hectic which has been taking my code time ^^;; )

discomrade:
No problem :) This one is big and has the mentioned caveats, so definitely best not to rush it.

discomrade:
Marked as draft to avoid this being merged as-is:
- The current code is capable of malicious DoS abuse and does not yet have sane default settings to mitigate this risk.
- If #1153 is merged, this will require significant refactoring. This code should be rewritten and resubmitted after that pull request is accepted.

PR #1105, opened 2024-03-11
Optimizing random

yls4:
Using RAND() function to make shuffle work faster.  I don't know what other databases are used and if they also have RAND().

### DIFF
--- a/core/imageboard/image.php
+++ b/core/imageboard/image.php
@@ -217,6 +217,24 @@ class Image implements \ArrayAccess
             return null;
         }
     }
+	
+	public static function by_random_multi(array $tags=[], int $limit_range=1): array
+    {
+        $max = Image::count_images($tags);
+        if ($max < 1) {
+            return null;
+        }		// From Issue #22 - opened by HungryFeline on May 30, 2011.
+        if ($limit_range > 0 && $max > $limit_range) {
+            $max = $limit_range;
+        }
+        $rand = mt_rand(0, $max-1);
+        $set = Search::find_images($rand, $limit_range, $tags, true);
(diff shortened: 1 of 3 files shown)


discomrade:
> I don't know what other databases are used and if they also have RAND().

As listed on the [Install documentation](https://github.com/shish/shimmie2/wiki/Install), Shimmie currently supports PostgreSQL 14+ / MySQL 8.0+ / SQLite 3.37+.

MySQL has RAND(). PostgreSQL and SQLite don't have RAND(), instead they have RANDOM(). The code will need to check which type of database it is to give the correct random function name. Something like:
```php
if ($database->get_driver_id() == DatabaseDriverID::MYSQL) {
    $order = "RAND()";
} else {
    $order = "RANDOM()";
}
```

I also noticed all the formatting and static analysis tests are failing too. Use [these tools to test your code](https://github.com/shish/shimmie2/blob/main/.github/CONTRIBUTING.md) :)

core/imageboard/image.php
```diff
@@ -217,6 +217,24 @@ public static function by_random(array $tags = [], int $limit_range = 0): ?Image
...
+        if ($limit_range > 0 && $max > $limit_range) {
+            $max = $limit_range;
+        }
+        $rand = mt_rand(0, $max-1);
+        $set = Search::find_images($rand, $limit_range, $tags, true);
+        if (count($set) > 0) {
+            return $set;
+        } else {
+            return null;
+        }
```
shish, line 236: Just `return $set;` seems correct here - if there are no results, then an empty array indicates "no results", where `null` is more like "the concept of results does not exist"

PR #1114, opened 2024-03-17
Minor improvements to tag category editor

thisisjoesGH:
Pulling this from a custom theme I have used for a while. adds a color swatch next to the hex code to show the current color of each category. also changes the category editor to use a color input instead of a text input to match the new category creation form

Screenshot_20240317_123749

I noticed the preview for the color input was too short to be visible so added an explicit height

color input before:
Screenshot_20240317_123843

color input after:
Screenshot_20240317_123850

### DIFF
--- a/ext/tag_categories/style.css
+++ b/ext/tag_categories/style.css
@@ -33,6 +33,10 @@
 	padding:0.2rem 0.6rem;
 }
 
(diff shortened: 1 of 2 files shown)


shish:
That all looks pretty great, thank you <3

PR #1119, opened 2024-03-22
[notes] Round note dimensions to int

### LINKED ISSUE #1118: Note creation fails when image is scaled down to fit window
1. Enable Notes extension
2. View a large image which gets downscaled to fit the browser window.
3. Attempt to add, move and save a note.

No error.

**What actually happened?**
Front end gives an alert that the note failed to be created.
Backend error: `PHP Fatal error:  Uncaught TypeError: Shimmie2\Notes::add_history(): Argument #4 ($noteX1) must be of type int, float given, called in /var/www/html/shimmie2/ext/notes/main.php on line 267 and defined in /var/www/html/shimmie2/ext/notes/main.php:390
`

discomrade:
Fixes #1118

### DIFF
--- a/ext/notes/script.js
+++ b/ext/notes/script.js
@@ -124,10 +124,10 @@ function renderEditor(noteDiv, note) {
 	function _commit() {
 		noteDiv.classList.remove("dragging");
 		dragStart = null;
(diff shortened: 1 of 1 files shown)


shish:
Looks good, thank you <3

PR #1120, opened 2024-03-23
Statistics page

### LINKED ISSUE #837: Booru statistics page
Gelbooru 0.1 has an stats page and it contains various rankings about user interaction, see for example: https://gravityfalls.booru.org/stats/

The page I want should have:
Top 10 taggers, Top 10 commenters, Top 10 image posters
Top 100 of each previously mentioned stat
Each top should mention place, amount and username. (just like Gelbooru)

Top 10 forum posters and Top 10 favoriters are optional.
Time settings are not needed, I think the stats should only be of all time.
Pie chart is not needed.

I think it would give an incentive to users to interact more with the site.

discomrade:
Closes #837

Future commits will add metrics for the notes and forum extensions.

### DIFF
--- a/ext/statistics/main.php
+++ b/ext/statistics/main.php
@@ -1,0 +1,147 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Shimmie2;
+
+class Statistics extends Extension
+{
+    /** @var StatisticsTheme */
+    protected Themelet $theme;
+
+    public function onPageRequest(PageRequestEvent $event): void
+    {
+        global $config, $page;
+        if ($event->page_matches("stats") || $event->page_matches("stats/100")) {
+            $base_href = get_base_href();
+            $sitename = $config->get_string(SetupConfig::TITLE);
+            $theme_name = $config->get_string(SetupConfig::THEME);
+
+            $limit = 10;
+            if ($event->page_matches("stats/100")) {
+                $limit = 100;
+            }
+
+            if (Extension::is_enabled(TagHistoryInfo::KEY)) {
+                $tag_tally = $this->get_tag_stats();
+                arsort($tag_tally, $flags = SORT_NUMERIC);
+                $tag_table = $this->theme->build_table($tag_tally, "Taggers", "Top $limit taggers", $limit);
+            } else {
+                $tag_table = null;
+            }
+
+            $upload_tally = [];
+            foreach ($this->get_upload_stats() as $name) {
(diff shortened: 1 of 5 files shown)


ext/statistics/main.php
```diff
@@ -0,0 +1,147 @@
...
+        return $database->get_col('SELECT users.name FROM comments INNER JOIN users ON users.id = comments.owner_id WHERE comments.owner_id <> 1;');
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function get_favorite_stats(): array
+    {
+        global $database;
+        // Returns the username of each comment, as an array. Excludes Anonymous
```
shish, line 144: s/comment/favourite/?

ext/statistics/main.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * @return array<string, int>
+     */
+    private function get_tag_stats(): array
+    {
+        global $database;
+        // Returns the username and tags from each tag history entry. Excludes Anonymous
+        $tag_stats = $database->get_all('SELECT users.name,tag_histories.tags,tag_histories.image_id FROM tag_histories INNER JOIN users ON users.id = tag_histories.user_id WHERE tag_histories.user_id <> 1;');
```
shish, line 94: In some rare cases anon's ID can be something other than 1 -- `$config->get_int("anon_id")` should always get the correct ID

ext/statistics/theme.php
```diff
@@ -0,0 +1,84 @@
...
+        }
+        if ($favorite_table) {
+            $html->appendChild($favorite_table);
+        }
+
+        $page->set_title(html_escape("Stats"));
+        $page->set_heading(html_escape("Stats - Top $limit"));
+        $page->add_block(new NavBlock());
+        $page->add_block(new Block("Stats", $html, "main", 20));
+    }
```
shish, line 40: MicroHTML itself checks for and ignores `null` elements, so this could be simplified to something like
```
$html = emptyHTML(
    $tag_table,
    $upload_table,
    $comment_table,
    $favorite_table,
);
```

shish (commented):
A couple of nits but looking lovely <3

discomrade:
Well-spotted, the anon_id one was a bit silly of me, I should have kept looking for a getter. I spotted the hardcoded `anon_id = 1` code in `comment/theme.php` so I'll fix that up too :)

PR #1140, opened 2024-04-26
Improvements to notes, tag histories, source histories

discomrade:
First commit adds the notes pages to the navbars. These were inaccessible before.

Second commit orders SQL requests by id instead of date, since date is not necessarily unique (especially if imported data from another booru software).


shish:
> Second commit orders SQL requests by id instead of date, since date is not necessarily unique

Could these use `ORDER BY date DESC, id DESC` so that date takes priority and ID is a tiebreaker? Date feels more "correct" to me (like, if an admin decided to go into the database and change the date column of an entry for some reason, they would expect it to show up in a different order), but yeah consistent ordering even with duplicate dates is nice 🤔

PR #1141, opened 2024-04-26
[user] fix typo in password check

discomrade:
In `user_admin/create_other`, would just send pass1 twice.


shish:
I think this was done with the intent of admins only needing to specify a password once (on the basis that admins are better a typing, and if they make a typo they have the power to fix it)... but then the UI has always had two textboxes, so yeah that's definitely inconsistent. Accepting this PR to make it consistent, and if admins complain that double-entering a password is too much work we can remove it from both the frontend and backend later /o/

PR #1144, opened 2024-04-28
Various captcha changes

discomrade:
- Adds captcha to the image upload form (either disabled, enabled for anon uploads only, or enabled for everyone)
- generalises captcha code instead of hardcoding the options for captcha on anonymous comments
- make the captcha for the signup page an option
- enable signup and anon comment captchas by default

### DIFF
--- a/core/captcha.php
+++ b/core/captcha.php
@@ -10,7 +10,7 @@ namespace Shimmie2;
 
 use ReCaptcha\ReCaptcha;
 
-function captcha_get_html(): string
+function captcha_get_html(bool $anon_only): string
 {
     global $config, $user;
 
@@ -19,7 +19,7 @@ function captcha_get_html(): string
     }
 
     $captcha = "";
-    if ($user->is_anonymous() && $config->get_bool("comment_captcha")) {
+    if (!$anon_only || $user->is_anonymous()) {
         $r_publickey = $config->get_string("api_recaptcha_pubkey");
         if (!empty($r_publickey)) {
             $captcha = "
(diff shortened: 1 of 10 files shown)


shish:
Longer term I'd love to use User Classes for everything and get rid of the special cases for `is_anonymous` and `is_logged_in` - this PR is definitely a step forwards compared to where we are today, but I wonder if it is adding more dependencies on the older system that will need to be undone later 🤔

(though without a GUI editor, the user class system currently isn't very user friendly, so I can't confidently say "definitely use that system and avoid putting user-behaviour settings on the board config screen" ^^)

Specifically I'm imagining having something like `$user->can(Permissions::SKIP_UPLOAD_CAPTCHA)`... I'mma think for a bit and see if I can come up with a way to make it user-friendly _and_ future-proof; if nothing comes to mind soon I'll merge this user-friendly approach and worry about the future in the future :)

discomrade:
I think you're right about this needing to be reworked later. Bypassing each captcha is better done at the permissions level (some boorus might only want staff to bypass, while others may want only anonymous users to be challenged and only for certain actions like commenting).

I think a GUI permissions editor is a good priority which would allow a lot more flexibility and remove assumptions. I am fine delaying and polishing this pull request until the permissions editor is in progress.

discomrade:
Will revise this pull code once #1149 is merged, because the new captcha bypass permissions will then involve a database upgrade.
Will also try adding an on-focus trigger for the comment captcha to avoid unnecessary loading on every post page.

shish:
Captchas got redone in a way that I think encompasses all of this
- captchas are now provided by extensions (with current options being "none" or "recaptcha", but it _should_ be trivial to add any others)
- captchas can be skipped for certain user classes based on permissions (eg by default users and admins will skip the comment-captcha, but anons will see it)
- captchas added to login (though by default everybody can skip it - you'll need to remove this permission from anons if you don't want them to skip it)
- captchas added to uploads

PR #1149, opened 2024-05-08
[permissions] move permissions to database, add GUI permissions editor

### LINKED ISSUE #1139: GUI Permission Editor
[Making a proper feature request for this suggestion:](https://github.com/shish/shimmie2/pull/1051#issuecomment-1956415249)

> at some point I'd love to have a GUI permission editor, which might work as a form of documentation so one can see the effects of enabling or disabling each flag 🤔

**Describe the solution you'd like**
A config page where the permissions of each role can be configured by administrators using checkboxes.

Special roles like admin should have their permissions visible with checkboxes greyed out. If a role has a parent role, this should be added using a dropdown box, so the user understands which parent role must be edited to remove an inherited permission, or can change the parent role.

In the future, this should be able to add new roles or delete unused roles.

discomrade:
Closes #1139

This does not handle existing user-defined classes (e.g. data/config/user-classes.conf.php) so I suggest the next release mention that users should remove from any custom user classes before upgrading.

This pull also implies changes to the wiki: https://github.com/shish/shimmie2/wiki/Advanced-Config


shish:
So this PR has been haunting me for months because I've wanted to get it merged, but something has felt off, but I wasn't able to say _what_ -- turns out my worries were nearly all in my own code that this builds on top of :laughing: So now that I've gone and redone how permissions are defined (putting them in per-extension files, rather than one central list), I'm now working on merging this on top of that /o/

One noteworthy change being that I'm thinking of making the permissions table more like `permissions(user_class, permission_id, allowed)` rather than having a separate column-per-permission -- the new foundations allow for adding and removing permissions when extensions are enabled or disabled, and doing a database schema change in each case sounds scary >.>;;

shish:
New permissions system merged, and done in a way that allows extensions to add new classes (with default-classes, old-style `user-classes.conf.php` classes and database-based classes currently supported) /o/

discomrade:
> New permissions system merged, and done in a way that allows extensions to add new classes (with default-classes, old-style `user-classes.conf.php` classes and database-based classes currently supported) /o/

Great news! My booru had this draft version implemented and file-based perms weren't an option (main admin doesn't have filesystem access) so the new perms overhaul was the last blocker before we can now hop back in line with the dev branch. \o\

PR #1157, opened 2024-06-03
[terms] add terms and conditions gate

discomrade:
This extension replaces pages with a terms and conditions message until they accept it (or login). The message can be customised e.g. to include an 18+ age limit.

It also adds the cookie to test requests to prevent them getting this page.

image


shish:
That all looks excellent, thank you <3

PR #1159, opened 2024-06-05
[devcontainer] get UID/GID from index.php for smoother container-based development

shish:
[devcontainer] get UID/GID from index.php for smoother container-based development


shish:
Gah, this makes things work on linux (where files in the container are mounted as the user account), buy breaks mac (where files are mounted as root)...

shish:
Actually this doesn't even work on linux, I'm not sure what I was thinking here

PR #1175, opened 2024-06-10
[blur] blur thumbs with explicit rating

discomrade:
This allows thumbs with certain ratings to be blurred until hovered over (or long-tapped on a touchscreen). This is made for boorus where explicit content should be available, but blurred out by default.

image

The admin can set the global default level to blur (e.g. Explicit, Questionable & Explicit, none) and user accounts can override it in their settings. Anon users receive the global default.

The second test function (`testRatingBlurGlobalConfig`) creates and deletes a test account. This is to simulate an anon user loading the global config. If there is a better way to do this, I can try and improve it.

### DIFF
--- a/core/basethemelet.php
+++ b/core/basethemelet.php
@@ -69,6 +69,12 @@ class BaseThemelet
                 $custom_classes .= "shm-thumb-has_child ";
             }
         }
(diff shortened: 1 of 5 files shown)


shish:
Tested and this all seems to work nicely, code looks good, thank you :D

PR #1184, opened 2024-06-18
[setup] have ConfigSaveEvent save the config

shish:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/shish/shimmie2/pull/1184).
* #1185
* __->__ #1184

### DIFF
--- a/ext/setup/main.php
+++ b/ext/setup/main.php
@@ -344,7 +344,6 @@ class Setup extends Extension
             $this->theme->display_page($page, $panel);
         } elseif ($event->page_matches("setup/save", method: "POST", permission: Permissions::CHANGE_SETTING)) {
             send_event(new ConfigSaveEvent($config, $event->POST));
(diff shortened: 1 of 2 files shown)


ext/setup/main.php
```diff
@@ -407,11 +406,8 @@ public function onConfigSave(ConfigSaveEvent $event): void
                 }
             }
         }
+        $config->save();
         log_warning("setup", "Configuration updated");
-        foreach (\Safe\glob("data/cache/*.css") as $css_cache) {
-            unlink($css_cache);
-        }
-        log_warning("setup", "Cache cleared");
```
shish, line 414: These files haven't lived here for a while, and cache-clearing shouldn't be needed since the theme name is now included as part of the hash filename

PR #1185, opened 2024-06-18
[core] use a dedicated CSRF salt instead of DATABASE_DSN

shish:


core/user.php
```diff
@@ -251,8 +251,14 @@ public function get_avatar_url(): ?string
     public function get_auth_token(): string
     {
         global $config;
-        $salt = DATABASE_DSN;
+
+        $salt = $config->get_string("csrf_salt");
+        if (is_null($salt)) {
+            $salt = bin2hex(random_bytes(32));
+            send_event(new ConfigSaveEvent($config, ["_type_csrf_salt" => "string", "_config_csrf_salt" => $salt]));
```
shish, line 258: This is a bit ugly, with ConfigSaveEvent specifically taking in the POST-formatted (ie, `string => string`) inputs. I feel like it should take a typed `["csrf_salt"=>$salt]` (ie `string => string/bool/int/array`) as input, and then the existing use-cases should use a POST-to-typed-array helper function... (that can be done as a follow-up though)

shish: Actually ConfigSaveEvent is a really inefficient way to do this, just `$config->set_string()` has an automatic save included

shish:
Actually, storing this in the database isn't great either

PR #1193, opened 2024-06-19
[core] function to generate a persistent salt

shish:
To be used in the next diffs


shish:
Actually, having the salt in the database means putting both parts of a two-part key in the same place, which isn't great

PR #1214, opened 2024-07-22
[bbcode] allow relative image links

discomrade:
This is useful for CSP when a site is hosted on multiple domains, such as a mirror.


shish:
Merged this by hand since I introduced a conflict ^^

PR #1238, opened 2024-09-01
Generate summary_large_image preview for images

BadMannersXYZ:
This adds extra `twitter:` and `og:` meta tags for posts (most importantly, `<meta property="twitter:card" content="summary_large_image">`), and changes the thumbnail link to the image link, in order to display full-sized images in embeds. For example, this is what a post looks like on Discord:

image


shish:
That all seems sensible, thank you ^__^

PR #1239, opened 2024-09-03
Fix a typo in a cache key

Mjokfox:
Message cache was not updating when reading a pm, found the issue to be a mistyped cache key


shish:
Thanks for spotting that :D

PR #1251, opened 2024-09-06
[docker] Build image for amd64 and arm64

BadMannersXYZ:
This modifies the Github Workflow steps by using Docker's official images for building to multiple architectures. I tried to keep parity with the current format, but I also added some new (optional) parameters for envvars that allow the user to choose a different image name and registry.

In order to test this, I created [a non-fork repo](https://github.com/BadMannersXYZ/shimmie2-test) (by default, Github doesn't allow forks to share secrets to workflows). This has [built an image](https://hub.docker.com/layers/badmanners/shimmie2/latest/images/sha256-066fcf3a1cd455e53d926ecc8c2a07ea00ad24f4bf370b0401ec3a46432a6de4?context=repo) that I've tested on my Raspberry Pi, and it seems to work perfectly, although I haven't tested the corresponding amd64 image yet.

### DIFF
--- a/.github/get-tags.py
+++ b/.github/get-tags.py
@@ -7,14 +7,17 @@ branch = check_output(["git", "branch", "--show-current"], text=True).strip()
 describe = check_output(["git", "describe", "--tags"], text=True).strip()
 tag = describe.split("-")[0][1:]
 a, b, c = tag.split(".")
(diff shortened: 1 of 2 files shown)


shish:
The code looks good, let's merge and see if it works in practice :eyes:

PR #1269, opened 2024-09-07
[WIP] limit=X meta-tag, see #1153

shish:
[WIP] limit=X meta-tag, see #1153

This seems close to working, except that "page size" and "total number of search results" are two distinct concepts, and SQL LIMIT is used for page size already, so it doesn't work to also use it for total number of search results...


sanmadjack:
As I stated in #1153 I have an implementation that covers these concerns. I'll get that together once I've update with the changes in the other ticket.

PR #1301, opened 2024-10-14
Improve text truncation speed

socram8888:
I was running 2.10.6 and noticed after updating to latest version that the booru was running way, way slower. I tracked it down to the truncate function, which was struggling with the thousands of words in the recent comments.

This PR severely improves the performance:

### Before patch

```
$ time curl http://192.168.0.245/booru/index.php?q=post/list -o/dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  103k    0  103k    0     0  20205      0 --:--:--  0:00:05 --:--:-- 26273

real    0m5.258s
user    0m0.000s
sys     0m0.031s
```

### After patch

```
$ time curl http://192.168.0.245/booru/index.php?q=post/list -o/dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  103k    0  103k    0     0  77359      0 --:--:--  0:00:01 --:--:-- 77317

real    0m1.413s
user    0m0.000s
sys     0m0.047s
```

The reason this is ultimately faster is that 1. the UTF-8 decoding is only executed once (vs multiple times as current for mb_strlen, mb_strrpos, mb_substr...), and 2. by calling mb_substr before, we will guarantee the length comparison will be bound to max ~50 characters (as opposed to strlen(a) != strlen(b) which will have to process all bytes in a and b to check their length).

### DIFF
--- a/core/polyfills.php
+++ b/core/polyfills.php
@@ -420,23 +420,38 @@ function clamp(int $val, ?int $min = null, ?int $max = null): int
 function truncate(string $string, int $limit, string $break = " ", string $pad = "..."): string
 {
     $e = "UTF-8";
-    $strlen = mb_strlen($string, $e);
     $padlen = mb_strlen($pad, $e);
     assert($limit > $padlen, "Can't truncate to a length less than the padding length");
 
-    // if string is shorter or equal to limit, leave it alone
-    if ($strlen <= $limit) {
+    /*
+     * Truncate tentatively, and then check if the lengths stayed the same.
+     *
+     * This approach is faster than calling mb_strlen and checking against the limit, as mb_strlen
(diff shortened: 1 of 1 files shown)


shish:
This does seem 4x faster, but FYI, if you install the mbstring library (eg php8.2-mbstring on debian), that'll be 1000x faster (I didn't realise how slow the polyfills were, maybe we should put more emphasis on having mbstring installed...)

```
0.69887208938599 - truncate() with polyfill
0.17476892471313 - truncate2() with polyfill
0.00040171194076 - truncate() with mbstring
0.00000390911102 - truncate2() with mbstring
```

PR #1314, opened 2024-12-03
Fix Favorites Bug

NottyNoz:
The id of the user was determining the post number that was being updated rather than the post's id. This only affected the `favorites>n` search query which is why it went unnoticed for 13 years.

sql statements for those interested
correct posts that may not have any favorites but whose favorite count was incorrectly updated
`UPDATE images SET favorites=(SELECT COUNT(*) FROM user_favorites WHERE image_id=id) WHERE favorites>0;`

set the correct favorite count for all favorited posts
`UPDATE images SET favorites=(SELECT COUNT(*) FROM user_favorites WHERE image_id=id) WHERE id IN (SELECT image_id FROM user_favorites);`


shish:
Nicely caught, thank you :D

PR #1358, opened 2024-12-14
define ShimmiePHPUnitTestCase second, so that phpstan sees it properly

shish:


shish:
Gah, this seems to work differently in CI vs IDE :/

PR #1372, opened 2024-12-30
Fix ouroboros

socram8888:
8c22cb1fb65c91adcb7f16c8cd59002693e17d83 broke the Ouroboros API. These two commits fix it.

The first one prevents a deprecation warning to be thrown, as $options should be zero when not used, not null.

The second one fixes it by avoiding the page_matches call entirely. Now it will also not set the page type unless something actually matches, so unimplemented methods just use Shimmie's default "no handler found" page instead of a blank empty page.

### DIFF
--- a/ext/ouroboros_api/main.php
+++ b/ext/ouroboros_api/main.php
@@ -151,10 +151,7 @@ class OuroborosPost extends _SafeOuroborosImage
             );
         }
         if (array_key_exists('description', $post)) {
-            $this->description = filter_var_ex(
-                $post['description'],
-                FILTER_SANITIZE_STRING
-            );
+            $this->description = $post['description'];
         }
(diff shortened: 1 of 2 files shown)


ext/ouroboros_api/main.php
```diff
@@ -151,10 +151,7 @@ public function __construct(array $post)
             );
         }
         if (array_key_exists('description', $post)) {
-            $this->description = filter_var_ex(
-                $post['description'],
-                FILTER_SANITIZE_STRING
-            );
+            $this->description = $post['description'];
```
shish, line 154: this feels like it _would_ be a security concern, if anything was reading `description` and expecting it to be safe -- but as far as I can see nothing is reading `description`? (and if at some point in the future we want to read `description`, it should be sanitised-on-output like everything else)

PR #1374, opened 2024-12-31
post id avatars

Mjokfox:
Wasn't a fan of 'gravatar', an external way to give users avatars, knowing e621 has the ability to set a post as your avatar, i wanted the same, and others might like it too so hence a pr.

since it uses user_config, it has to get the user config every time for the owner of every comment, using user_config::get_for_user($id) seemed slow as it sends a quite large event every time, so it directly gets the config in User::get_avatar_post_link(), but i am not entirely sure if its safe to do it in this way. The url or full html might also be cacheable to increase efficiency a bit thinking about it now.

### DIFF
--- a/core/user.php
+++ b/core/user.php
@@ -238,7 +238,11 @@ class User
     {
         // FIXME: configurable
         global $config;
-        if ($config->get_string("avatar_host") === "gravatar") {
+        $avatar_host = $config->get_string("avatar_host");
+        if ($avatar_host === "post"){
+            return $this->get_avatar_post_link();
+        }
+        elseif ($avatar_host === "gravatar") {
             if (!empty($this->email)) {
                 $hash = md5(strtolower($this->email));
                 $s = $config->get_string("avatar_gravatar_size");
@@ -248,6 +252,21 @@ class User
                 return "https://www.gravatar.com/avatar/$hash.jpg?s=$s&d=$d&r=$r&cacheBreak=$cb";
             }
         }
+        return null;
+    }
+
+    public function get_avatar_post_link() : ?string {
+        global $database;
+        $user_config = new DatabaseConfig($database, "user_config", "user_id", (string)$this->id);
+        $id = $user_config->get_int("avatar_post_id");
+        if ($id === 0) {
+            return null;
+        }
+        $image = Image::by_id($id);
+        if ($image) {
+            return $image->get_thumb_link();
+        }
+        $user_config->set_int("avatar_post_id", 0);
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


discomrade:
Hi @Mjokfox, it looks like the formatting test didn't pass. The `main` branch is locked down so it can't accept pull requests that don't pass the tests.

https://github.com/shish/shimmie2/actions/runs/12563091233/job/35024471162?pr=1374

The tests can be run locally, see [CONTRIBUTING.md](https://github.com/shish/shimmie2/blob/main/.github/CONTRIBUTING.md) for examples.

I'd like to see this added, already looks great with those foxes.

Mjokfox:
Thanks for the feedback! Not familiar yet with such checks, fixed it now :3

ext/user/main.php
```diff
@@ -409,6 +409,7 @@ public function onSetupBuilding(SetupBuildingEvent $event): void
 
         $hosts = [
             "None" => "none",
+            "Post id" => "post",
```
shish, line 412: Capitalise ID

core/user.php
```diff
@@ -251,6 +254,22 @@ public function get_avatar_url(): ?string
...
+        $user_config = new DatabaseConfig($database, "user_config", "user_id", (string)$this->id);
+        $id = $user_config->get_int("avatar_post_id", 0);
+        if ($id === 0) {
+            return null;
+        }
+        $image = Image::by_id($id);
+        if ($image) {
+            return $image->get_thumb_link();
+        }
+        $user_config->set_int("avatar_post_id", 0);
```
shish, line 269: I'm guessing that this line is here so that when a user's avatar-post is deleted, then the next time we try to find their avatar, then it will fail to find the image, then fall back to setting it to 0, and then _next_ time it will shortcut and immediately return null? If that is the intention, then I would do `$user_config->delete()` rather than set to zero

shish (commented):
A couple of very minor nits but yes this looks good to me :)

shish:
(Also apologies for the delay with getting to this, I've been on a yak-shaving adventure, trying to resuscitate an abandoned library that shimmie depends upon >.>;; )

core/user.php
Mjokfox: correct, changed

Mjokfox:
Alright, added your suggestions.

I do still have the idea of adding a way to crop and resize the avatars using CSS, so the actual image does not need to change, while giving the user the ability to have their avatar be only part of an image. Basically store a "zoom" and "center position" in the db. I haven't tried any of that yet but I'd be willing to try and make it work if that is of any interest here as well?

shish:
What's here now looks like a useful and stable milestone so I'mma merge as-is~

(Adding cropping sounds like it'd be nice - though I wouldn't count myself as "interested" per se, since the galleries I'm involved with already have a lot of user-investment in the gravatar approach and users are resistant to change :sweat_smile: But yeah it seems like a good thing for new installs to use)

Issue #1390, opened 2025-02-06
parse_shorthand_int() cannot parse negative values.


Mjokfox:
I have found that parse_shorthand_int() cant parse negative values, as it is used in ConfigSaveEvent::postToSettings() for the "int" type, i would expect an integer to also be able to be negative, unlike unsigned of course. 

would it be possible to make it also parse negative values? I see that the default return is -1, which probably is an indicator for if the parsing failed right now, which might as well be null instead then.
The only change would be adding "-?" to the regex on line 462 of polyfills.php, like so

`'/^(-?[\d\.]+)([tgmk])?b?$/i'`

PR #1391, opened 2025-02-07
[user] Post avatars with cropping

Mjokfox:
- Add a button in post controls to set the image as avatar.
- Add a page to set the avatar cropping easily.
This page should make it easy for the user to crop their avatar to their liking, they can drag it around and zoom in using the slider. And if they do not have JavaScript enabled, it can be manually changed on the user options page.

All the cropping is done by CSS and does not require any changes to be made to the actual images, nor copying and resizing.

I also included the parse_shorthand_int change i mentioned in https://github.com/shish/shimmie2/issues/1390 because it does rely on that right now. A workaround would be to parse some values as string instead, then cast it to integer afterwards manually but i don't see issue in having parse_shorthand_int be able to parse negative values as well.

### DIFF
--- a/ext/user/main.php
+++ b/ext/user/main.php
@@ -339,23 +339,19 @@ class UserPage extends Extension
         }
 
         if ($event->page_matches("set_avatar/{image_id}", method: "GET", permission: Permissions::CHANGE_USER_SETTING)) {
-            if (!$user->is_anonymous()) {
-                $image_id = int_escape($event->get_arg('image_id'));
-                $this->theme->display_avatar_edit_page($page, $image_id);
-            }
+            $image_id = int_escape($event->get_arg('image_id'));
+            $this->theme->display_avatar_edit_page($page, $image_id);
         }
 
         if ($event->page_matches("save_avatar", method: "POST", permission: Permissions::CHANGE_USER_SETTING)) {
-            if (!$user->is_anonymous()) {
-                $settings = ConfigSaveEvent::postToSettings($event->POST);
-                send_event(new ConfigSaveEvent($user_config, $settings));
-                $page->flash("Image set as avatar");
-                $page->set_mode(PageMode::REDIRECT);
-                if (key_exists(AvatarConfig::POST_AVATAR_ID, $settings) && is_int($settings[AvatarConfig::POST_AVATAR_ID])) {
-                    $page->set_redirect(make_link("post/view/".$settings[AvatarConfig::POST_AVATAR_ID]));
-                } else {
-                    $page->set_redirect(make_link("user_config"));
-                }
+            $settings = ConfigSaveEvent::postToSettings($event->POST);
+            send_event(new ConfigSaveEvent($user_config, $settings));
+            $page->flash("Image set as avatar");
+            $page->set_mode(PageMode::REDIRECT);
+            if (key_exists(AvatarConfig::POST_AVATAR_ID, $settings) && is_int($settings[AvatarConfig::POST_AVATAR_ID])) {
+                $page->set_redirect(make_link("post/view/".$settings[AvatarConfig::POST_AVATAR_ID]));
+            } else {
+                $page->set_redirect(make_link("user_config"));
             }
         }
     }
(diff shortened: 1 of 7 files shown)


Mjokfox:
Actually adding
```css
.avatar-editor .avatar-container {
	border: 2px solid var(--text);
}
```
to ext/user/style.css might be nice just realized, to show the borders of the image while cropping it on /set_avatar

ext/user/main.php
```diff
@@ -326,6 +331,33 @@ public function onPageRequest(PageRequestEvent $event): void
...
         }
+
+        if ($event->page_matches("set_avatar/{image_id}", method: "POST", permission: Permissions::CHANGE_USER_SETTING)) {
+            $image_id = int_escape($event->get_arg('image_id'));
+            $page->set_mode(PageMode::REDIRECT);
+            $page->set_redirect(make_link("set_avatar/$image_id"));
+        }
+
+        if ($event->page_matches("set_avatar/{image_id}", method: "GET", permission: Permissions::CHANGE_USER_SETTING)) {
+            if (!$user->is_anonymous()) {
```
shish, line 342: Why the extra check for anons? I would think just checking for "change user setting" permission should already exclude anons (unless for some weird reason the admin has decided to give anons permission to edit their settings, but in that case it seems like they should be allowed to)

shish:
One minor question but generally looks good :)

(Seeing so much avatar-handling code in `ext/user` where any given site will only be using gravatar _or_ posts (or neither?) makes me think that we should probably send a `GetAvatarEvent($user)` and then either `ext/avatar_gravatar` or `ext/avatar_post` would handle it... that could be done in the future though)

Mjokfox:
Fixed that.
Yeah i started to get the feeling that a specific extension for avatars might be something to consider as well.

shish:
Merging as-is since this seems like a sensible milestone /o/

Are you interested in splitting `user` into `user` / `avatar_gravatar` / `avatar_post`? I'm happy to either leave that to you or add it to my own to-do list, just checking so that we don't both try to do it at the same time :smile:

Mjokfox:
Oh i'll look into it :3
Do you think it makes more sense to have both as extensions, and whichever is enabled is the provider by default, making them mutually exclusive to each other. Or still leave an option in setup which seems more convoluted to me.
I see that you can set `$conflicts` in `ExtensionInfo` so the first option makes more sense to me.

shish:
My first thought was to allow both being enabled, with `BuildAvatarEvent::setAvatar($html)` overwriting the output if two extensions call it, which would be confusing but not break anything

Then when you point out that `$conflicts` exists, my second thought is "that's a better idea, let's do that"

But then my third thought is "Actually why _not_ allow both at once"?

- BuildAvatarEvent has an `$html` property which defaults to empty
  - it has a `setAvatar($html)` function which sets the html, and then also sets `$this->stop_processing = true;` so that no other extensions will see the event
- Give the "post avatars" extension a higher priority so it gets called first
  - if the user has a post ID set, then call `$event->setAvatar($html);`
  - if they don't have a post ID set, then do nothing
- If the `BuildAvatarEvent` is still being processed, then it gets passed on to the Gravatar extension
  - if the user has an email set, then call `$event->setAvatar($html);`
  - if they don't have an email set, then do nothing

That way the avatar generation process would try post id (if enabled), fallback to gravatar (if enabled), fallback to empty - which also has the nice side effect of allowing existing gravatar-heavy sites to migrate over to post avatars gradually

shish:
(Also incidentally this seems like a nice opportunity to switch over from string-concatenation to `MicroHTML`, in order to make a whole category of typo-based security holes impossible)

Mjokfox:
That sounds like a good solution, so i'm thinking to make it similar to how formatting works. I'm now adding `BuildAvatarEvent` and `AvatarExtension` to `core/Event.php` and `core/Extension.php` respectively to make a similar interface for this. 

Also, it is kinda odd that theres a whole `TextFormattingEvent` and `FormatterExtension` when BBcode is the only one that implements it by default, but i've made use of it so i won't complain.

shish:
Yeah, I keep thinking that it'd be nice to add a markdown formatter since markdown appears to have won the "human-readable-ish formatting" war - but since the sites I run have already heavily-invested in bbcode, and I can't see a sensible way to run both in parallel, I wouldn't be able to use it :sweat_smile:

Mjokfox:
I actually made a markdown formatter on my fork, and all in java script so `wysiwyg` is possible. I have to say its quite messy and not perfectly complete in terms of markdown functionality, but has been running fine on my site. See here: https://github.com/Mjokfox/shimmie2/tree/Fork/ext/markdown
Having both bbcode and markdown at once *should* be possible i do think, but i disabled BBcode because i was not familiar with it

PR #1392, opened 2025-02-07
[style] various word wrapping changes

discomrade:
Prevents words being broken unnecessarily in places like comments and wiki pages, plus general clean up of wrapping CSS rules.

- Convert alias property `word-wrap` to `overflow-wrap` for consistency
- Use `overflow-wrap: anywhere;` instead of `word-wrap: break-all;` to prevent words being broken unnecessarily
- Use `overflow-wrap: anywhere;` instead of `overflow-wrap: break-word;` to work better with flex
- Move H1, H2, H3 wrapping rule in default theme to static css rule on all themes

All the existing word wrapping CSS rules become `overflow-wrap: anywhere;`. I am not confident with simply replacing all these rules with one global root declaration, as the default wrapping behavior was chosen to avoid any chance of overflowing data being hidden.

### DIFF
--- a/ext/bbcode/style.css
+++ b/ext/bbcode/style.css
@@ -1,5 +1,5 @@
 .bbcode {
-	word-break: break-all;
+	overflow-wrap: anywhere;
 }
 .bbcode BLOCKQUOTE {
 	border: 1px solid black;
--- a/ext/comment/style.css
+++ b/ext/comment/style.css
@@ -2,7 +2,7 @@
 .comment {
 	text-align: left;
 	position: relative;
-	word-wrap: break-word;
+	overflow-wrap: anywhere;
 }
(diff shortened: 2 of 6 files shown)


shish:
Sticking "wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww" into various textboxes around my test site, I'm not seeing any overflow problems, so LGTM :)

PR #1405, opened 2025-02-07
[avatars] Separate gravatar and pavatar in their own extensions

Mjokfox:
Remove the avatar handling from `user.php` and `ext/user`, and bring them over to their own extensions.

see https://github.com/shish/shimmie2/pull/1391#issuecomment-2643390172

Pavatars are cached by default for one minute, to help with large amounts of comments from the same user for example.

Both pavatar and gravatar can be enabled at once, with pavatar taking priority if a valid image is set. Else gravatar will try if the email is set. If both fail, a notice is shown on the user's profile for themselves

I have kept `get_avatar_post_url()` for pavatar, but it is never used, and wasn't used before either, there was the function:
```
#[Field(name: "avatar_url")]
    public function get_avatar_url(): ?string
```
Which also was never used, but that #[Field] is for the cli version right? I'm not sure about that, but it could be returned by adding another event just like the html one.


core/extension.php
```diff
@@ -290,6 +292,27 @@ abstract public function format(string $text): string;
...
+    {
+        global $cache;
+        $html = $this->avatar_html($event->user);
+        if ($html) {
+            $event->setAvatar($html);
+            $event->stop_processing = true;
+        }
+    }
+
+    abstract public function avatar_html(User $user): HTMLElement|false;
```
shish, line 312: `null` feels more accurate than `false` here

shish (approved):
This looks excellent :D


> #[Field]

Those annotations are used to automatically generate the GraphQL API; I'm not currently aware of any users of that API (except for my own personal experimental client) so I think it's fine to remove it if it no longer makes sense

Mjokfox:
The default css is still under `ext/user` but i think that is fine because its the same for gravatar and pavatar anyways, and user is always active. But now there *could* be the possibility to have different styling for both

shish:
That all sounds sensible - I'mma merge now so that I can start making use of it on my testbed, though if you think of more improvements you're welcome to keep sending PRs :)

PR #1447, opened 2025-02-11
php 8.2 is deprecated

shish:


shish:
8.3 doesn't seem to have huge benefits, so maybe one more release with 8.2 compatibility :thinking:

PR #1511, opened 2025-02-22
[user] fix the redirect setting to work properly again

Mjokfox:
Broken in commit a833f5fa5c7d390fef28968196e4a8ef435da15e
Where the setting before relied on `0` being interpreted as false in php, now the value is a string, and thus will always evaluate as true unless empty string which is not possible with the default given, nor the options given in `user/config.php`.
before:
`if ($config->get_int("user_loginshowprofile", 0))`


shish:
Thank you :)

PR #1516, opened 2025-02-26
Fix navlink bug in extensions

ermnick:
Using a leading slash in the navlinks causes the post list to fail to load with the following error:
```php
Internal Error
Message: make_link(/post/list/in%3Atrash/1): page cannot start with a slash

Version: 2.12.0-alpha (on 8.4.4)

Stack Trace:

#0 /app/core/urls.php(20): Shimmie2\make_link()
#1 /app/themes/danbooru2/page.class.php(142): Shimmie2\Link->make_link()
#2 /app/themes/danbooru2/page.class.php(108): Shimmie2\Danbooru2Page->navlinks()
#3 /app/core/page.php(565): Shimmie2\Danbooru2Page->body_html()
#4 /app/core/page.php(289): Shimmie2\Page->render()
#5 /app/index.php(106): Shimmie2\Page->display()
#6 /app/index.php(153): Shimmie2\main()
#7 {main}
```
Only affected extensions seems to be trash and approval.

### DIFF
--- a/ext/approval/main.php
+++ b/ext/approval/main.php
@@ -93,7 +93,7 @@ class Approval extends Extension
         global $user;
         if ($event->parent == "posts") {
             if ($user->can(ApprovalPermission::APPROVE_IMAGE)) {
-                $event->add_nav_link("posts_unapproved", new Link('/post/list/approved%3Ano/1'), "Pending Approval", null, 60);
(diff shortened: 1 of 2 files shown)


shish:
Thank you for catching this, I shall also add a new unit test to specifically cover navlinks which aren't used in the default theme /o/

PR #1553, opened 2025-03-06
couple small bugfixes, and bandaid subnav fix

Mjokfox:
With the new usage of  `$link` being a string, it always has a / in front of it, so it never matches `$query`, which does not have a slash in front. So trimming the slash away fixes it, but i feel like is a bandaid fix.

Besides that, the tag help link was pointing to the incorrect page.
And `CommandBuilder->execute()` always failed for me with `Cannot access uninitialized non-nullable property CommandBuilder::$output by reference `


shish:
Thank you <3

Issue #1423, opened 2025-02-10
Wiki functionality improving


BraVoRg:
Hello! 
I propose improving the wiki. Currently, the wiki is just a small system of descriptions for tags, but overall it has great potential and interesting functionality. I have made some improvements, and you can view them on the test page of my website: https://broneatlas.ru/wiki/тест
This means I have the ability to insert links, embed images, and make them clickable (this functionality might already exist, but I couldn't find it in Shimmie version 2.11.3).

 have two questions:

How to create a user role as a wiki administrator so that my friends can write articles?
How to find a list of all edited wiki articles, because currently, I can't figure out which articles I have?

shish:
> This means I have the ability to insert links, embed images, and make them clickable

All these should be doable with bbcode already


> How to create a user role as a wiki administrator so that my friends can write articles?

https://github.com/shish/shimmie2/wiki/Custom-User-Classes


> How to find a list of all edited wiki articles

That isn't there at the moment, but it does sound useful 🤔

BraVoRg:
>All these should be doable with bbcode already
I tested now how bbcode works. I couldn't find way to resize the image. I write: [img=100x50] and... this is not work. But another functions of bbcode still works! Thank you! 

Also a lot of thanks for custom classes! I will do moderator class)

shish:
>  Also will make script which will find all wiki articles. Maybe need to do a variable when you create an article its name will write in.

Yeah, since we already have special pages for `wiki:sidebar` and `wiki:default`, I was thinking something like `wiki:list` could have a special case where the content is auto-generated 🤔

shish:
> I couldn't find way to resize the image

That isn't there yet, and checking [1] / [2] I don't see any standard way to do it 🤔 

[1] https://www.phpbb.com/community/help/bbcode
[2] https://en.wikipedia.org/wiki/BBCode

BraVoRg:
> Yeah, since we already have special pages for `wiki:sidebar` and `wiki:default`, I was thinking something like `wiki:list` could have a special case where the content is auto-generated 🤔

Yes, this will be very helpful when you accidetally create a mistake article. 

>That isn't there yet, and checking [1] / [2] I don't see any standard way to do it 🤔

I back my way to add images, using just urls? 40% width and centered text align and also added a way to add by bbcode. They were conflicting before, now there s two parallel ways to add images. 

Also not working colors, tables and text aligning in bbcode, trying to fix it.

shish:
[...]

PR #1559, opened 2025-03-12
[wiki] List all wiki pages

Mjokfox:
I think mentioned like this in #1423, simply a wiki page that lists all existing pages


shish:
Excellent :D Thank you ^_^

PR #1578, opened 2025-03-14
[various] Re-add the missing subnavs

Mjokfox:
Both subnavs on wiki and the account page were missing, adding them back with the `$matches`  parameter


shish:
Whoops, sorry and thank you <3

PR #1731, opened 2025-04-10
[statistics] fix html string in tag history field

discomrade:
Fixes a minor oversight introduced in commit 384100cbc5a9f053c674c0da0aa70d88de2e93c6


shish:
Thank you <3

PR #1741, opened 2025-04-11
Update VideoCodec.php

Miyuutsu:
Adds to VideoCodec.php ensuring that AVC and HEVC are correctly handled. Essentially fixes a regression since these worked fine with the extension in the last stable release. I do not know if it's necessary or useful to have the H264 and HEVC handlers so I have left those in as they shouldn't cause any problems by existing.


shish:
Looks good, thank you :)

PR #1751, opened 2025-04-12
Fix Submit button for Edit Tag Category / Submit button for New Tag Category

gabbah-draws:
Found this bug on latest changes on main, so here's a PR that restores the functionality of editing/adding new tag categories!

Repro steps:
- Clone shimmie2 from main branch
- Run shimmie2
- Install Tag Categories extension
- Navigate to the "Tag Categories" page (tags/categories)
- Try to edit an existing tag from the defaults
- Button will be unresponsive due to a "SyntaxError" on the "Edit" button.
- Try to create a new Tag Category
- Button will be unresponsive as the "Submit" button is outside of the new Tag Category form

### DIFF
--- a/ext/tag_categories/theme.php
+++ b/ext/tag_categories/theme.php
@@ -62,7 +62,7 @@ class TagCategoriesTheme extends Themelet
                     BUTTON([
                         "class" => "tc_edit",
                         "type" => "button",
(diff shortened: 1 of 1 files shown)


shish:
Thank you <3

PR #1752, opened 2025-04-12
[forum] fix threadExists() returning false always for psql

Mjokfox:
sqlite returns an integer with this db query, however psql returns a boolean. So with the strict equals, it returns false always. Not tested with mysql


shish:
Thank you for catching this - While reviewing the change I found a bunch of other issues (and a complete lack of unit tests which would have caught any of them), fixed and tested in #1753

PR #1767, opened 2025-04-19
[ext] Update auto_tagger functionality and rating stabilty

Miyuutsu:
I'm not sure if this type of drastic change is acceptable or not, but I added AND as well as ANDNOT functionality to the auto_tagger extension and rewrote small parts of the theme to handle it. The update to rating was necessary because somehow something was breaking somewhere and this seems to be much more stable if anything isn't perfect.

### DIFF
--- a/ext/auto_tagger/main.php
+++ b/ext/auto_tagger/main.php
@@ -240,54 +240,59 @@ final class AutoTagger extends Extension
     {
         global $database;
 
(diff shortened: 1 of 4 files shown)


Miyuutsu:
Sorry for all the draft/ready to check there. I'm not overly used to GitHub yet.

Miyuutsu:
Should probably not be merged. I ran into some issues with the new auto-tag.

PR #1795, opened 2025-05-12
[ext] Image description extension

xifize:
This is a simple extension to add descriptions or artist commentary to image posts. I decided to put the descriptions in a separate DB table instead of the images table because they're stored as TEXT without a fixed size.


shish:
Looks good, thank you - I shall merge and bring it up to date with the other recent changes /o/

shish:
Actually minor thing that I realise immediately after clicking merge - I'm also gonna rename to "Post Description" because it works for more than just images (For ancient compatibility reasons things are called "images" inside the database, but they can actually be any type of file, and I'm trying to consistently call them "posts" where possible)

PR #1814, opened 2025-05-29
[handle_pixel] improved click-to-zoom and resizing

Mjokfox:
Always was annoyed by click to zoom not actually zooming to the location clicked, so tried to improve on it

### DIFF
--- a/ext/handle_pixel/main.php
+++ b/ext/handle_pixel/main.php
@@ -67,12 +67,11 @@ final class PixelFileHandler extends DataHandlerExtension
             $event->add_part(FORM(
                 SELECT(
                     ["class" => "shm-zoomer"],
-                    OPTION(["value" => "full"], "Full Size"),
-                    OPTION(["value" => "width"], "Fit Width"),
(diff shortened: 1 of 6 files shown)


shish:
I like this behaviour - it seems incompatible with the Notes extension though :cry:  (When zooming in, notes are scaled up appropriately - but they are also placed over the top of the page instead of staying inside the scrollable area, and they don't pan when the image pans)

Mjokfox:
Adding this eventlistener in `notes/script.js` makes the notes follow again when panning.
`noteImage.parentElement.addEventListener("scroll", renderNotes);`
or for a bit lighter approach on the client side, using `scrollend` so it does not trigger every frame of a scroll
`noteImage.parentElement.addEventListener("scrollend", renderNotes);`

however the overlaying issue seems a bit more difficult to tackle, i will look into it

shish:
@Mjokfox heads up that #1910 moves this into a separate zoom_to_click extension (with the original behaviour moved into zoom_to_fit, so a site can pick which zoom style to use, or have both, or neither)

PR #1843, opened 2025-07-11
[lowercase] allow admin to force lowercase tags

discomrade:
- Forces new tags into lowercase before adding to database
- Instructs admins how to change existing tags
- Styles tag fields to show input in lowercase, hinting to users that tags are treated as lowercase

Due to the simple function of the extension and how complicated it would be to handle it in existing unit testing, I believe it is reasonable to disable during unit testing.

The CSS reuses the `autocomplete_tags` class to find tag fields in the HTML. This is admittedly a hack, and therefore this is a draft pull request until this is either accepted or a better solution is decided on. My concerns are that I haven't checked if all tag fields are `autocomplete_tags` fields, and if so, should `autocomplete_tags` simply be replaced with `tags`?

### DIFF
--- a/core/ImageBoard/Tag.php
+++ b/core/ImageBoard/Tag.php
@@ -33,6 +33,11 @@ final class Tag
         );
         if (empty($id)) {
             // a new tag
+
+            // lowercase extension ruins unit tests, so disable during testing
+            if (!defined("UNITTEST") && ForceLowercaseInfo::is_enabled()) {
+                $tag = mb_strtolower($tag);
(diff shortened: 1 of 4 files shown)


discomrade:
> my gut says to have a Config option on `post_tags` rather than having a separate extension

I originally tried to have a Config option on `system`, but I didn't know how to make the CSS rule apply conditionally, without over-complicating it.

shish:
Added as an option in #1879 <3

PR #1964, opened 2025-09-02
[system] fix referer link could be null

Mjokfox:
If not enough extensions were enabled, or a user did not have enough permissions, the `/system` page would have the redirect be `null`, throwing an error


Mjokfox:
@HayateLaTech, get a fresh install of shimmie2, do the basic setup, make admin account etc.
Then **log out** and go to the page `/system`, which throws an internal error. This is because there is no page the user is allowed to view with the basic permission. But if for example tag_history is enabled, and the user had that permission, no error would be thrown.

shish:
I keep looking at this and thinking it would be better to avoid generating the link in the first place if there's nothing for it to point to, but it's been a month and I haven't figured out how to do that in an elegant way short of redoing the whole navigation system (which could be a good idea, but is a lot of work), so merging as-is for now /o/

PR #1972, opened 2025-09-17
 Auto "AI" tagging via Automatic 1111 and WD14 

Saetron:
This is a basic plugin which sends the image over to a running automatic 1111 instance with the tagger extension installed to get automated tags and ratings for the image. Threshold and Model can be changed via the Shimmie2 interface. Requires a running Automatic1111 instance, so I dont know if you want to add it, but it saved me quite alot of work for getting a new shimmie2 instance running. (Danbooru has a similar feature).

### DIFF
--- a/ext/automatic1111_tagger/main.php
+++ b/ext/automatic1111_tagger/main.php
@@ -8,21 +8,26 @@ class Automatic1111Tagger extends Extension
 {
     public const KEY = "automatic1111_tagger";
 
-    public function onImageBlockBuilding(ImageBlockBuildingEvent $event)
+    public function onImageBlockBuilding(ImageBlockBuildingEvent $event): void
     {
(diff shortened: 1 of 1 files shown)


shish:
I like this idea /o/

FYI you can run the automated tests locally with `composer check` (or `composer format` / `composer test` / `composer analyze` to run an individual subset of tests)

I'd like to get it running myself to test with, can you confirm that the two specific bits of software I need are:
- https://github.com/AUTOMATIC1111/stable-diffusion-webui
- https://github.com/picobyte/stable-diffusion-webui-wd14-tagger

Saetron:
Here is an example of it working. Get Rating pretty much does the same as interrogate, but only pushes the rating of the image and no tags.

example

PR #2018, opened 2025-11-25
Autocomplete Fix

Noztik:
When you have a list of autocompletes and then start typing, the autocomplete list visually gets filtered to match your new text while an autocomplete request is sent to the api. However, that filtered autocomplete list isn't actually the real list because it is still using the old list so if you navigate to the first item in the filtered list, you are actually selecting the first item in the old list. This is only a problem if you navigate after you change your search query and before the api response arrives.

https://github.com/user-attachments/assets/e68479d4-6670-4298-96e8-b6eaf3c1a29e


shish:
Looks good, thank you <3

PR #2043, opened 2026-01-02
[Forum] Cleaning up and getting the forum extension code up to date

Mjokfox:
Annoyed by the forum not having events like comments, so with heavy inspiration from the comment extension, rewrote most of the forum extension to be cleaner and more up to date with the rest of the codebase, without making major changes to the functionality.

I haven't touched the theme much, which i think also could use some upgrades.

### On altering functionality:
Something like a reply button like comments have would be a great addition, something that shouldn't be too difficult to add with some js.

On my fork, I've made it possible for users to edit [image] comments and forum posts already, not sure if you are also interested in that for the main repo? Feel free to let me know!

### DIFF
--- a/ext/ban_words/main.php
+++ b/ext/ban_words/main.php
@@ -15,6 +15,20 @@ final class BanWords extends Extension
         }
     }
 
+    public function onForumThreadPosting(ForumThreadPostingEvent $event): void
+    {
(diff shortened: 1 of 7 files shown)


Mjokfox:
for the comment / forum post editing, do you think its important to make sure it works without javascript? Its definitely possible with just css i've found, just more finicky

PR #2048, opened 2026-01-07
[core] more specific type name, so we can check non-string content later

shish:


Mjokfox:
This looks nice to me, makes everything more generic which is good
only thing i noticed is the event->context is a bit odd, it should definitely use constants, instead of hardcoded strings
`if ($event->context === "tag" || $event->context === "source") {`
it should definitely be
`if ($event->context === PostTags::context || $event->context === PostSource::context) {`
and so on

PR #2058, opened 2026-01-19
Change to devcontainers + new data attr for thumbs

luanalatte:
Bundled these 2 together since I don't see a reason to create a separate PR for the 2nd small commit.

The biggest one here is the change to the devcontainer. I'm looking for feedback on that. Basically:

- Specified docker.io source for podman compatibility.
- Moved the postStartCommand to the Dockerfile itself so that vscode doesn't get stuck on the "Configuring" stage.
- Removed duplicate apt update and install lines.

The other change is just adding a data attribute to thumbnails when the video flag is set. I use it to show a blue border around those in my theme.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,6 +8,8 @@
 		"target": "devcontainer"
 	},
 
+	"overrideCommand": false,
+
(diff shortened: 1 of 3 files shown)


shish:
Looks like there is some precedent which made this easy :O (`image/webp` vs `image/webp; lossless=true`)

Issue #2062, opened 2026-01-22
EventListener attribute


luanalatte:
What do we think about a change to the event bus making use of PHP 8 attributes?

Something like this:

```php
#[EventListener]
public function methodCanHaveAnyNameNow(PageRequestEvent $event): void {}
```

then we use reflection to get the event type from the first argument or like this if preferred
```php
#[EventListener(PageRequestEvent::class)]
public function methodCanHaveAnyNameNow(PageRequestEvent $event): void {}
```

or error out if none is provided.

We could also have per-listener priority instead of per-class, since the priority could be an argument in the attribute tag.

```php
#[EventListener] // Default priority 50
public function onImageDeletion(ImageDeletionEvent $event): void {}

#[EventListener(priority: 10)]
public function onPageRequest(PageRequestEvent $event): void {}
```

or even have a single extension listen for the same event two times. This could be useful to hook very early into a block building event for example, then apply some finishing touches once the rest of the extensions have added their stuff.

```php
#[EventListener(priority: 1)]
public function onPageNavBuildingStart(PageNavBuildingEvent $event): void {}

#[EventListener(priority: 99)]
public function onPageNavBuildingEnd(PageNavBuildingEvent $event): void {}
```

I already have these solved locally, would love to go ahead with a PR if it's appreciated.

EDIT: I'd also add theme's `page.class.php` to the event listener checks. This way themes can add their own listener without having to create a custom extension :)

PR #2063, opened 2026-01-23
Event Listener change

luanalatte:
Suggested implementation for #2062 . Completely replaces the old method, so we would have to re-add it below the new one if we want to keep backwards compat for 2.12. Also, there is some room for quick optimizations here besides the obvious caching

### DIFF
--- a/core/Events/EventBus.php
+++ b/core/Events/EventBus.php
@@ -68,9 +68,7 @@ final class EventBus
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


shish:
I have YOLO-merged this because the idea seems solid enough to give it more battle testing and fix forwards ^^ (Merged manually rather than clicking the github merge button as there were a bunch of conflicts that needed fixing)

PR #2065, opened 2026-01-26
Add HEVC codec support for MP4 video

Saetron:
Currently upload fails for videos if the mime type is identified as HEVC, as H.265 and HEVC are pretty much the same, this shouldn't break anything. Tested it on my install and it works as expected, even with file transforming extensions.

### DIFF
--- a/ext/handle_video/main.php
+++ b/ext/handle_video/main.php
@@ -15,6 +15,7 @@ final class VideoFileHandler extends DataHandlerExtension
         MimeType::MP4_VIDEO => [
             VideoCodec::H264,
             VideoCodec::H265,
+            VideoCodec::HEVC,
         ],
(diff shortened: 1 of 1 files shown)


shish:
This confuses me because as I understand it H.265 and HEVC aren't pretty much they same, they are _literally_ the same :sweat_smile: 

As far as I can tell h265/hevc support was added previously based on "what should work in theory", and only a few different combinations of container and codec were actually tested in practice >.> #2074 switches to use "hevc" everywhere as that appears to work in practice (And also deduplicates some of the code, and also adds AV1 as a valid webm codec)

If I've misunderstood the situation and there are actually two distinct file formats which need to be supported in two ways, would you be able to give an example file for each that I can test with? ^^

PR #2153, opened 2026-04-04
Fix the type problem of cron upload

He1l0-520:
The `cronuploader` encountered two errors. It was discovered that the new `path` data type was not handled in the `generate_image_queue function`, and the issue was resolved by using explicit type conversion.


shish:
Thank you <3

PR #2154, opened 2026-04-06
Enable cron_uploader to extract tags from image-matching JSON files

He1l0-520:
Extend cron_uploader to extract tags from JSON files with the same name as images, enabling batch image uploads from the server with tags automatically applied.

Also, add json to `is_skippable_file` so the json file don't get moved to `fail_to_upload` folder.

(Sorry for the empty commit, it is generated while updating fork, I am still learning how to get rid of it QWQ)

### DIFF
--- a/ext/automatic1111_tagger/info.php
+++ b/ext/automatic1111_tagger/info.php
@@ -9,9 +9,9 @@ final class Automatic1111TaggerInfo extends ExtensionInfo
     public const KEY = "automatic1111_tagger";
 
     public string $name = "AI Tagger";
-    public array $authors = ["Saetron" => "https://github.com/Saetron/Shimmie2-AI-Tagger"];
+    public array $authors = ["Saetron" => "https://github.com/Saetron"];
     public string $license = self::LICENSE_MIT;
     public string $description = "Connects to the Automatic1111 API to interrogate images and suggest tags.";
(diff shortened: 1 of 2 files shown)


shish:
I like this idea, I'm mostly wondering if there's a way to make "upload from directory (via cron)", "upload from directory (via admin interface)" and "upload from directory (via zip file)" share some of their code, since the zip upload already has support for uploading files with _all_ metadata fields (tags, source, rating, etc) and it'd be nice to support that :thinking:

He1l0-520:
I didn't realize that ZIP uploading support json file tag handling, I will try to reuse that part of code. Thank U for the hint. ❤️

Btw, can I take a look at the error message of `Tests / Format` failure? Does that failure matter or not? 😭

图片

shish:
I suspect that getting the several different types of folder-upload to share code will be quite a bit more work, but you're welcome to try ^^;; But getting `cron_upload` to use `PostInfoSetEvent` feels like it should be doable and have most of the benefits (ie, making it possible to set all post metadata in one go)

FWIW the .zip uploader expects a single `export.json` containing metadata for all files, but I think having a .json per file actually makes more sense... (Also right now the .zip uploader pulls data directly from the .zip, but it probably makes more sense to extract to a temporary folder and then use a shared "upload folder" function)

You can automatically format the code by running `composer format`

PR #2156, opened 2026-04-12
Fix cbz files not loading due to SCRIPT function escaping the quotes

dgtal-git:
Fixes an issue where the comic reader would not load the file and just keep showing the loading icon, due to the MicroHTML SCRIPT function URL escaping the quotes inside the string.
image


shish:
I feel like this extension needs quite a bit more polish, but this does make it not-broken at least ^^

PR #2189, opened 2026-04-30
[Pools] fix pools/list not showing anything

Mjokfox:
The ilike pattern got the wildcards removed since bc28b4537c7169b5b171ce49a691c88a9f81e368 . So simply add them back, but now the :search and title are still not LOWER(), which probably would be nice to add back as well, but im not sure how that should be done together with the SCORE_ILIKE() function, especially since with sqlite it already adds the LOWER, but not for mysql and pgsql

### DIFF
--- a/ext/pools/main.php
+++ b/ext/pools/main.php
@@ -577,7 +577,7 @@ final class Pools extends Extension
 		", [
             "l" => $poolsPerPage,
             "o" => $pageNumber * $poolsPerPage,
-            "search" => $search,
+            "search" => "%$search%",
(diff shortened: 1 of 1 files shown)


shish:
`SCORE_ILIKE` gets translated into whatever the database needs to do a case-insensitive-pattern-match:

- sqlite turns it into `LOWER(x) LIKE LOWER(y)`
- mysql's `LIKE` is case-insensitive by default
- postgres uses `ILIKE` (explicit case-insensitive version of `LIKE`)

PR #2201, opened 2026-05-13
[ext] Fix notes extension

Sakimotor:
The current version of the extension, taken from the master branch, would crash upon creating/saving a note due to some numbers being read as floats rather than ints.

There was also an issue with an SQL request on the note_histories table being made with an erroneous variable name ("enable" instead of "note_enable")


shish:
Thank you for catching that <3

PR #2207, opened 2026-05-26
Add configurable OAuth2 login extension

### LINKED ISSUE #912: Oauth2 support for account creation and login
**Is your feature request related to a problem? Please describe.**
A lot of sites that have boorus often have them as an extension to an already existing site (such as a forum or a fediverse instance), unfortunately this means that you need to have two accounts and keep track of two passwords, the convenient thing would be to use a single account on their main service and then use oauth for each other service they host, including a booru. 
This also allows people to control who they let into their booru easier, if it is for example a private community. Because even with an account approval type system, it needs someone to moderate it. With the current options that becomes manually generating accounts and handing them out on request. But with oauth, the admins only need to approve users on their main service, and then they can sign up to the other services that the site provides easily on their own.

**Describe the solution you'd like**
The idea would be to have an oauth button on the login page which prompts authentication with the administrators specified site (in addition to, or replacing the standard login with username system), it would be convenient to have multiple oauth options so that a user can log in from more than one site as well, such as if a booru is shared between a few associated sites.

What would likely be the best and lowest maintainance option would be to follow the format that [writefreely](https://github.com/writefreely/writefreely) uses for it's "[Generic Oauth](https://writefreely.org/docs/latest/admin/config#Generic_OAuth)" and allow the user to customise on their own each part of the oauth, so that sites using non standard names for the data that it provides back to the client can be used too (such as pleroma providing "fqn" instead of "email"). This way you're not spending the time trying to support multiple sites individually, but instead are allowing the user to customise it as they need, and solve a problem that happens with some oauth implementations where sites send back different data names than some software expects. Gitea for example will not let you generate an account unless the oauth server provides back the "email" field.

**Describe alternatives you've considered**
There isn't really an alternative as far as I am aware, I don't know of any booru software doing something like this.

**Additional context**
If more explaination or clarification is needed, I am open to giving it.

nexiumbiz-debug:
Fixes #912.

This adds an OAuth2 login extension with two supported deployment shapes:

- trusted reverse-proxy login for oauth2-proxy/nginx-style setups, enabled only when explicitly configured
- direct OAuth2 authorization-code login using administrator-configured provider endpoints

The trusted reverse-proxy mode covers the confirmed bounty scope from #912: it uses configurable username and email headers, maps the trusted upstream identity to an existing or newly-created Shimmie user, and documents the security boundary that Shimmie must not be reachable directly by clients when header trust is enabled.

The direct OAuth2 client path includes:

- configurable provider name, client credentials, authorization/token/userinfo endpoints, scopes, and profile field mapping
- a short-lived state cookie for callback validation
- JSON token/userinfo requests, verified-email enforcement by default, and existing-account matching by email
- automatic Shimmie user creation with normalized unique usernames when no existing email match is found
- a sidebar login block plus focused tests for login links, provider redirect, callback state rejection, user creation, unverified-email rejection, and trusted proxy header login

Verification run locally:

- PHP lint for `ext/oauth2_login` files
- `php vendor/bin/phpunit --no-coverage --filter "OAuth2LoginTest|UserPageTest"` (9 tests / 26 assertions)
- `php vendor/bin/php-cs-fixer check --diff --using-cache=no ext\oauth2_login`
- `php vendor/bin/phpstan analyse --memory-limit 1G --error-format=raw --no-progress`

Prepared with Codex assistance.

### DIFF
--- a/ext/oauth2_login/README.md
+++ b/ext/oauth2_login/README.md
@@ -60,6 +60,11 @@ Set these Shimmie board config values:
 - Proxy email header: `X-Forwarded-Email`
 - Auto-create OAuth2 users: enabled
 
+Some proxies expose the authenticated username as `REMOTE_USER` instead of an
+HTTP header. For example, Authelia deployments can set Proxy username header to
+`REMOTE_USER`; Shimmie reads that exact server variable instead of converting it
+to `HTTP_REMOTE_USER`.
+
 The following files can be dropped into a checkout of
 `shish/shimmie2-examples` next to the existing `docker-compose.yaml`. The
 values are intentionally local-development only.
--- a/ext/oauth2_login/README.md
+++ b/ext/oauth2_login/README.md
@@ -1,0 +1,233 @@
+# OAuth2 Login extension
+
+This extension supports two deployment shapes:
+
+- direct OAuth2 login, where Shimmie talks to the provider itself
+- trusted reverse-proxy login, where a fronting service such as
+  oauth2-proxy authenticates the request and passes a username/email header to
+  Shimmie
+
+For both modes, enable the `oauth2_login` extension first.
+
+## Direct OAuth2 client test with Keycloak
+
+Create a local Keycloak realm and confidential client:
+
+- realm: `shimmie`
+- client id: `shimmie-local`
+- client authentication: enabled
+- client secret: `local-dev-secret`
+- valid redirect URI: `http://localhost:4010/oauth2_login/callback`
+- web origin: `http://localhost:4010`
+- test user: `shimmie-admin`
+- test user email: `shimmie-admin@example.test`
+- mark the test user's email as verified
+
+Then set these Shimmie board config values:
+
+- Provider name: `Keycloak`
+- Client ID: `shimmie-local`
+- Client secret: `local-dev-secret`
+- Authorization URL:
+  `http://keycloak.localhost:8080/realms/shimmie/protocol/openid-connect/auth`
+- Token URL:
+  `http://keycloak.localhost:8080/realms/shimmie/protocol/openid-connect/token`
+- Userinfo URL:
+  `http://keycloak.localhost:8080/realms/shimmie/protocol/openid-connect/userinfo`
+- Scopes: `openid profile email`
+- Username fields: `preferred_username,name,email`
+- Email field: `email`
+- Email verified field: `email_verified`
+- Require verified email: enabled
+- Auto-create OAuth2 users: enabled
+
+Visit `http://localhost:4010/`, click "Log in with Keycloak", and sign in as
+the Keycloak test user.
+
+## Trusted reverse-proxy testbed
+
+This mode is useful when oauth2-proxy or another fronting service owns all
+OAuth2 redirects. Shimmie only trusts the identity headers set by that proxy.
+
+Only enable trusted-header mode when clients cannot reach Shimmie directly.
+The reverse proxy must clear user-supplied identity headers before it sets the
+trusted values.
+
+Set these Shimmie board config values:
+
+- Trust reverse proxy headers: enabled
+- Proxy username header: `X-Forwarded-User`
+- Proxy email header: `X-Forwarded-Email`
+- Auto-create OAuth2 users: enabled
+
+The following files can be dropped into a checkout of
+`shish/shimmie2-examples` next to the existing `docker-compose.yaml`. The
+values are intentionally local-development only.
+
+### docker-compose override
+
+```yaml
+services:
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.0
+    command: start-dev --import-realm --hostname=keycloak.localhost --hostname-strict=false
+    ports:
+      - "8080:8080"
+    environment:
+      KC_BOOTSTRAP_ADMIN_USERNAME: admin
+      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
+    volumes:
+      - ./oauth2-keycloak-realm.json:/opt/keycloak/data/import/shimmie-realm.json:ro
+    networks:
+      default:
+        aliases:
+          - keycloak.localhost
+
+  oauth2-proxy:
+    image: quay.io/oauth2-proxy/oauth2-proxy:v7.7.1
+    command:
+      - --provider=keycloak-oidc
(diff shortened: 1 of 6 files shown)


ext/oauth2_login/main.php
```diff
@@ -0,0 +1,445 @@
...
+        return $value === "" ? null : $value;
+    }
+
+    private function header_server_key(string $header): string
+    {
+        $header = strtoupper(str_replace("-", "_", trim($header)));
+        if ($header === "") {
+            return "";
+        }
+        if (str_starts_with($header, "HTTP_") || $header === "REMOTE_USER") {
```
MattDahEpic, line 110: This is an edge case if the header you are looking for *is* `REMOTE_USER` which is the case for Authelia. Documentation should include this edge case

ext/oauth2_login/main.php
```diff
@@ -0,0 +1,445 @@
...
+
+        $email = $this->configured_header_value(OAuth2LoginConfig::PROXY_EMAIL_HEADER) ?? "";
+        if ($email !== "" && !filter_var($email, FILTER_VALIDATE_EMAIL)) {
+            throw new InvalidInput("Trusted proxy returned an invalid email address");
+        }
+
+        $user = $this->find_or_create_proxy_user($username, $email);
+        send_event(new UserLoginEvent($user));
+        $user->set_login_cookie();
+        Log::debug("oauth2-login", "Logged in @$user->name from trusted reverse proxy headers");
```
MattDahEpic: This should be an info level log as normal logins are an info level log

shish:
I've never deployed oauth2 myself - do you have any guides or examples so that I can try and get this running on my own testbed install? (It's hard for me to do ongoing maintenance of code if I'm unable to run it on my own hardware ^^)

Bonus points if it's something that can be easily dropped into https://github.com/shish/shimmie2-examples/ - but even just some documentation or "this is the oauth provider I used and here's the settings which are known to work" would be a good start :3 (I see there's some docs included which explain how to connect shimmie to a provider, but I don't have a provider yet o.o)

nexiumbiz-debug:
Added a testbed guide in `ext/oauth2_login/README.md` for this.

It now covers two local paths:

- direct OAuth2 client mode using a local Keycloak realm/client/user, including the exact Shimmie config values
- trusted reverse-proxy mode using Keycloak + oauth2-proxy + nginx, with copyable `docker-compose` service snippets, `oauth2-nginx.conf`, and `oauth2-keycloak-realm.json` intended to be easy to drop into `shimmie2-examples`

I also added a pointer to the README from the extension info text. I validated the Markdown JSON example parses and `git diff --check` is clean. I couldn't run `php -l` locally because PHP is not available on this Windows PATH, but the only PHP change is the short docs pointer string.

nexiumbiz-debug:
I also opened a concrete `shimmie2-examples` follow-up for the drop-in testbed: https://github.com/shish/shimmie2-examples/pull/2

I updated this PR's README to match that example's `localhost:4050` reverse-proxy flow, so the docs and examples repo now line up.

nexiumbiz-debug:
Thanks, addressed both review notes in `7647b26`:

- changed trusted reverse-proxy login logging from debug to info
- documented the `REMOTE_USER`/Authelia edge case and added a focused regression test for that server variable path

Verification:
- `php -l ext/oauth2_login/main.php`
- `php -l ext/oauth2_login/test.php`
- `php vendor/bin/phpunit --no-coverage --filter OAuth2LoginTest` (7 tests / 14 assertions)
- `php vendor/bin/php-cs-fixer check --diff --using-cache=no ext\oauth2_login`
- `git diff --check`

ext/oauth2_login/info.php
```diff
@@ -0,0 +1,38 @@
...
+declare(strict_types=1);
+
+namespace Shimmie2;
+
+final class OAuth2LoginInfo extends ExtensionInfo
+{
+    public const KEY = "oauth2_login";
+
+    public string $name = "OAuth2 Login";
+    public array $authors = self::SHISH_AUTHOR;
```
shish, line 12: I'm not the author for this one 👀

ext/oauth2_login/info.php
```diff
@@ -0,0 +1,38 @@
...
+<pre><code>proxy_set_header X-Forwarded-User \"\";
+proxy_set_header X-Forwarded-Email \"\";
+auth_request /oauth2/auth;
+auth_request_set \$user \$upstream_http_x_auth_request_user;
+auth_request_set \$email \$upstream_http_x_auth_request_email;
+proxy_set_header X-Forwarded-User \$user;
+proxy_set_header X-Forwarded-Email \$email;</code></pre>
+
+See <code>ext/oauth2_login/README.md</code> for a local Keycloak and
+oauth2-proxy testbed that can be adapted into shimmie2-examples.";
```
shish, line 37: I'd link to the shimmie2-examples repo rather than having example configs embedded here

ext/oauth2_login/main.php
```diff
@@ -0,0 +1,445 @@
...
+            curl_setopt($ch, CURLOPT_HTTPHEADER, array_merge(
+                ["Accept: application/json", "Content-Type: application/x-www-form-urlencoded"],
+                $headers
+            ));
+        }
+
+        $response = curl_exec($ch);
+        $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
+        $error = curl_error($ch);
+        curl_close($ch);
```
shish, line 246: `curl_close()` is no-op since 8.0 and deprecated in 8.5

ext/oauth2_login/README.md
```diff
@@ -0,0 +1,236 @@
...
+- Auto-create OAuth2 users: enabled
+
+Some proxies expose the authenticated username as `REMOTE_USER` instead of an
+HTTP header. For example, Authelia deployments can set Proxy username header to
+`REMOTE_USER`; Shimmie reads that exact server variable instead of converting it
+to `HTTP_REMOTE_USER`.
+
+The following files can be dropped into a checkout of
+`shish/shimmie2-examples` next to the existing `docker-compose.yaml`. The
+values are intentionally local-development only.
```
shish, line 70: These blocks are all included in `shimmie2-examples` now, so I'd avoid duplicating them here

shish (commented):
I'm still fiddling and trying to get this working reliably - for some reason getting lots of internal server errors from oauth2-proxy, but then _sometimes_ it logs in ok, I've not figured out the pattern yet O_o

PR #2212, opened 2026-06-23
[index] Remove non-existent meta-tag

Mjokfox:
This meta-tag got added way back in 9b0edcf4499f1e63177e35b10b32d96fca76c83b but i couldn't find anything about an images.phash row, so i'm not sure why it got added in the first place?

### DIFF
--- a/ext/index/main.php
+++ b/ext/index/main.php
@@ -190,9 +190,6 @@ final class Index extends Extension
         } elseif ($matches = $event->matches("/^(hash|md5)[=:]([0-9a-fA-F]*)$/i")) {
             $hash = strtolower($matches[2]);
             $event->add_querylet(new Querylet('images.hash = :hash', ["hash" => $hash]));
-        } elseif ($matches = $event->matches("/^(phash)[=:]([0-9a-fA-F]*)$/i")) {
-            $phash = strtolower($matches[2]);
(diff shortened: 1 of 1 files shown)


shish:
There was some experimental perceptual-hash work where an external system was hashing images and populating that database column, which could then be searched in shimmie to find similar-looking images, but I think it got abandoned (and if somebody wants to revive it, the code should live in a perceptual-hash extension instead of the core), so yeah good to clean that up from here /o/

Mjokfox:
I have made such extension, two even for that matter.
One using a php library, https://github.com/jenssegers/imagehash to do perceptual hashing, t he library is a bit abandoned, and is quite slow to run, but of course easy to integrate. see https://github.com/Mjokfox/shimmie2/tree/main/ext/duplicate_detector

And one using openai's clip to get more like a "visual" hash that can correlate both similar image and text descriptions, so it doubles as a way to find similar images, and giving the option to have a text search query instead of tag only query. This one does need the model to run in the background at all times, and to be honest i made the extension very poorly, wanting to improve it at some point. see https://github.com/Mjokfox/shimmie2/tree/main/ext/reverse_image

I use both extensions together to give a popup warning on the upload page when someone tries to upload an image that appears to be a duplicate.

If the any seems appealing to be put in the main repo, i'd be glad to help integrate it properly

PR #2235, opened 2026-09-03
Add dark default theme

mesmerus:
I took the default theme, since I like it the best, and cloned it, changing the colors so its naturally dark. Figured I'd share it back; somebody else might appreciate it.


mesmerus:
@shish I'm not sure what I'm missing, but I can't seem to run the workflow tests. I tried running them in my fork as well, but there's no button to actually run them there. If you let me know what I need to do to get the tests to run, I do believe they'd all pass, given this just adds two files for the theme.

shish:
I think it just needed manual approval as you've not contributed to this repo before :eyes: 

(Also sorry for the delays, new job has been hectic so my open source projects have slowed down ^^; )

mesmerus:
Aha - I knew the main repo was protected, I thought the tests were supposed to run automatically on Sunday from something I'd read in the codebase somewhere; when it didn't happen (twice) I started looking for a way I was supposed to run the tests, even in the fork :joy_cat: 

No worries; was figuring I'd missed something, thanks for running those! I really like your default theme, I just figured it could do with a dark version. And if I type up such a change, why not share? :heart:

